PR #21, opened 2026-01-18
Replace std::ops::Not::not with is_false helper for improved readability

MatejGomboc:
## Summary

Addresses Clippy warnings from [code quality AI findings](https://github.com/embedded-society/altium-designer-mcp/security/quality/ai-findings) by replacing the obscure `std::ops::Not::not` trait method with a clearer helper function in `src/mcp/server.rs`.

## Changes

- Add `is_false(b: &bool) -> bool` helper function for use with serde's `skip_serializing_if` attribute
- Replace `std::ops::Not::not` with `is_false` in two locations:
  - `ToolCapabilities::list_changed` field
  - `ToolCallResult::is_error` field

The helper is a `const fn` and includes `#[allow(clippy::trivially_copy_pass_by_ref)]` since serde requires the `&T` signature.

## Impact

Code readability improvement only — no functional changes.

### DIFF
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -59,7 +59,8 @@ pub struct ToolCapabilities {
     pub list_changed: bool,
 }
 
-fn is_false(b: &bool) -> bool {
+#[allow(clippy::trivially_copy_pass_by_ref)] // serde requires &T signature
+const fn is_false(b: &bool) -> bool {
     !*b
 }
 
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -55,8 +55,12 @@ impl Default for ServerCapabilities {
 #[derive(Debug, Clone, Default, Serialize)]
 pub struct ToolCapabilities {
     /// Whether the tool list can change during the session.
-    #[serde(rename = "listChanged", skip_serializing_if = "std::ops::Not::not")]
+    #[serde(rename = "listChanged", skip_serializing_if = "is_false")]
     pub list_changed: bool,
+}
+
+fn is_false(b: &bool) -> bool {
+    !*b
 }
 
 /// Server information for initialisation response.
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -143,7 +143,7 @@ pub struct ToolCallResult {
     /// Content returned by the tool.
     pub content: Vec<ToolContent>,
     /// Whether the tool call resulted in an error.
-    #[serde(skip_serializing_if = "std::ops::Not::not")]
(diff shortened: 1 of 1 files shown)


MatejGomboc-Claude-MCP (changes requested):
## Code Review

Thanks for applying the Copilot Autofix suggestions! I've reviewed the changes and whilst the intent is good, there are some significant issues that need to be addressed before this can be merged.

### ✅ Good Changes

**1. `is_false` helper function** — Clean improvement. The named function is clearer than `std::ops::Not::not`.

**2. Inline closure `|b| !b`** — Simple and idiomatic fix.

**3. `get_library_reading_tools()` extraction** — Good direction for reducing function length.

### ❌ Critical Issues

**1. `parse_symbols_from_json()` is an incomplete placeholder**

The method currently returns:
```rust
Err("parse_symbols_from_json is not yet implemented".to_string())
```

This will cause `call_write_schlib()` to **always fail** with this error message. The actual parsing logic from the original function was never moved into this helper.

**2. Inconsistent refactoring in `call_write_schlib()`**

The code now calls `parse_symbols_from_json()` and assigns the result to `symbols`, but the subsequent code still uses `symbols_json` directly:
```rust
let new_names: Vec<&str> = symbols_json // ← Still using symbols_json, not symbols
```

The refactoring is incomplete — either the full parsing logic needs to be moved into the helper, or the call to it should be removed.

### Recommendation

This PR should either:
1. **Complete the refactoring** by moving the actual symbol parsing logic into `parse_symbols_from_json()`, or
2. **Revert the incomplete changes** (items 3 and 4) and keep only the first two fixes which are correct

I'd suggest closing this PR and applying only fixes 1 and 2 manually, then addressing the `too_many_lines` warnings in a separate, properly implemented PR.

MatejGomboc-Claude-MCP (dismissed):
✅ **Approved**

The PR has been cleaned up and now contains only the two working fixes:

1. **`is_false` helper** — Clean helper function for `skip_serializing_if` attribute
2. **Both `std::ops::Not::not` replacements** — Now consistently use the `is_false` helper

This is a safe, minimal refactoring that improves code readability without any functional changes. The broken incomplete refactoring (placeholder `parse_symbols_from_json()` and partial `get_library_reading_tools()` extraction) has been properly removed.

Good to merge.

MatejGomboc-Claude-MCP (approved):
LGTM! Clean refactoring that improves code readability.

The `is_false` helper is a good solution for serde's `skip_serializing_if` requirement — much clearer than using `std::ops::Not::not` directly. Nice use of `const fn` and the clippy annotation is appropriate given serde's signature constraints.

PR #25, opened 2026-01-18
Fix for Except block handles 'BaseException'

MatejGomboc:
Replace bare `except:` with `except Exception:` in `scripts/analyze_pcblib.py` to avoid catching `KeyboardInterrupt`, `SystemExit`, and other process-control exceptions.

This follows Python best practices and ensures the programme can be interrupted or exited normally, whilst still handling decode errors gracefully.


MatejGomboc-Claude-MCP (approved):
LGTM! Clean fix that follows Python best practices. Bare `except:` clauses are problematic because they catch everything including `KeyboardInterrupt` and `SystemExit`, which should propagate normally. Using `except Exception:` is the correct approach here.

PR #26, opened 2026-01-18
Fix for Empty except

MatejGomboc:
## Summary

Fixes CodeQL warning about empty `except` block in `parse_properties()` function.

## Changes

- Moves `return props` into the `try` block for explicit normal-flow return
- Adds explanatory comment in `except` block clarifying that malformed property data is intentionally ignored
- Explicitly returns `props` from the `except` block to satisfy CodeQL whilst preserving existing behaviour

## Impact

No functional change — the function continues to return any successfully parsed properties, gracefully handling malformed input. The code is now more explicit about its intent.

### DIFF
--- a/scripts/analyze_schlib.py
+++ b/scripts/analyze_schlib.py
@@ -21,9 +21,10 @@ def parse_properties(data: bytes) -> dict:
             if '=' in part:
                 key, value = part.split('=', 1)
                 props[key] = value
+        return props
     except Exception:
-        pass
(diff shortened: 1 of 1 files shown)


MatejGomboc-Claude-MCP (approved):
Good fix! The changes properly address the CodeQL warning whilst maintaining the existing graceful-degradation behaviour. The added comment makes the intent clear for future readers.

PR #33, opened 2026-01-18
Add StrokeFont enum for stroke text font type differentiation

MatejGomboc:
Implements stroke font support for Text primitives in PcbLib files:

- Add `StrokeFont` enum with `Default`, `SansSerif`, and `Serif` variants
- Add `stroke_font` field to `Text` struct  
- Parse stroke font ID from geometry block bytes 25-26
- Write stroke font ID to geometry block

## Description

This PR exposes the stroke font type differentiation for text primitives in PcbLib files. Altium Designer supports three stroke font variants (Default, Sans-Serif, and Serif) for vector text, and this change allows the MCP server to correctly read and write these font types.

The stroke font ID is stored as a `u16` at bytes 25-26 in the text geometry block:
- 0: Default
- 1: Sans-Serif  
- 2: Serif

The field is optional and only applies when `TextKind` is `Stroke`.

## Type of Change

## Standards Checklist

- [x] Primitive types (Pad, Track, Arc, etc.) maintain backwards compatibility
- [x] Altium file format compatibility is maintained
- [x] No breaking changes to existing MCP tool interfaces

## Testing

- [x] I have tested these changes locally
- [x] New and existing tests pass (`cargo test`)

## Code Quality

- [x] Documentation updated if needed
- [x] TODO.md updated to mark stroke font tasks as complete

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -88,12 +88,12 @@ This document tracks implementation gaps between the documented PcbLib format (`
 - [x] Parse text kind from geometry block (offset 1)
 - [x] Write text kind to geometry block (offset 1)
 
-### Stroke Font IDs - Not Exposed
+### Stroke Font IDs - Implemented ✓
 
-- [ ] Add `StrokeFont` enum: `Default`, `SansSerif`, `Serif`
-- [ ] Add `stroke_font: Option<StrokeFont>` field to `Text`
-- [ ] Parse stroke font ID (bytes 25-26 in geometry block)
-- [ ] Write stroke font ID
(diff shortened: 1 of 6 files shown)


MatejGomboc-Claude-MCP (approved):
Looks good! 👍

Clean implementation that follows the existing code patterns. The `StrokeFont` enum is well-documented, the parsing/writing logic correctly handles the byte offsets, and the optional field with `skip_serializing_if` is the right approach for backwards compatibility.

Nice work marking off the TODO items as well.

PR #46, opened 2026-01-24
Fix/3d step model persistence

MatejGomboc:
## Description

Fix 3D STEP model persistence and add SchLib batch parameter updates.

### 3D Model Persistence Fix

The `step_model` parameter in `write_pcblib` was being accepted but not persisted to the file. This PR implements the full roundtrip:

**Write path:**
- Convert `Model3D` to `ComponentBody` primitive
- Read STEP file from disk and create `EmbeddedModel`
- Generate GUID for the model and write to `/Library/Models/`

**Read path:**
- Parse model names from `/Library/Models/Data` stream
- Populate `model_3d` from `component_bodies` for backward compatibility

### SMD Pad Layer Fix

Fixed `Pad::smd()` to use `TopLayer` instead of `MultiLayer`. SMD pads should be on a single copper layer, not multi-layer which is for through-hole pads.

### SchLib Batch Parameter Updates

Extended `batch_update` tool to support SchLib files with a new `update_parameters` operation:
- Set parameter values across multiple symbols
- Optional regex filtering via `symbol_filter`
- `add_if_missing` option to create parameters if not present

## Type of Change

## Standards Checklist

- [x] Primitive types (Pad, Track, Arc, etc.) maintain backwards compatibility
- [x] Altium file format compatibility is maintained
- [x] No breaking changes to existing MCP tool interfaces

## Testing

- [x] I have tested these changes locally
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass (`cargo test`)

## Code Quality

- [x] CI checks pass (build, test, clippy, fmt)
- [x] Documentation updated if needed
- [x] CHANGELOG.md updated for user-facing changes

## New Dependencies

- `uuid` v1.11 - for generating GUIDs for 3D models
- `regex` v1.11 - for symbol filtering in batch updates

## Additional Notes

The `PcbLib::write()` signature changed from `&self` to `&mut self` to support the 3D model conversion. This is technically a breaking change for the Rust API but not for the MCP tool interface.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,9 @@ bitflags = { version = "2.6", features = ["serde"] }
 # Character encoding (for Windows-1252 in legacy Altium files)
 encoding_rs = "0.8"
 
+# UUID generation (for 3D model GUIDs)
+uuid = { version = "1.11", features = ["v4"] }
+
 [dev-dependencies]
 # Async testing utilities
 tokio-test = "0.4"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,9 @@ encoding_rs = "0.8"
 # UUID generation (for 3D model GUIDs)
 uuid = { version = "1.11", features = ["v4"] }
 
+# Regex for pattern matching (batch_update symbol filter)
+regex = "1.11"
+
 [dev-dependencies]
 # Async testing utilities
 tokio-test = "0.4"
--- a/README.md
+++ b/README.md
@@ -324,8 +324,8 @@ Returns:
 
 ### `batch_update`
 
-Perform batch updates across all components in an Altium PcbLib file. Supports updating track widths
-and renaming layers library-wide.
+Perform batch updates across all components in an Altium library file. Supports PcbLib
+operations (track width updates, layer renames) and SchLib operations (parameter updates).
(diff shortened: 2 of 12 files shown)


MatejGomboc-Claude-MCP (approved):
Excellent work! This PR delivers solid fixes and enhancements:

**3D Model Persistence** - The core bug is properly addressed with a clean bidirectional conversion between `Model3D` ↔ `ComponentBody`/`EmbeddedModel`. The implementation correctly handles GUID generation, STEP file embedding, and backward-compatible reading.

**SMD Pad Layer Fix** - Good catch. SMD pads on `MultiLayer` would indeed cause issues in Altium. `TopLayer` is the correct default.

**SchLib Batch Updates** - The `update_parameters` operation is a welcome addition. The regex filtering is particularly useful for large libraries.

**Tests** - The new `test_model_3d_persistence` and `test_model_3d_embedded_data_persistence` tests properly verify the roundtrip behaviour.

**Minor note:** The signature change from `&self` to `&mut self` on `PcbLib::write()` is technically a breaking change for downstream Rust consumers, but it's necessary for the 3D model preparation step and doesn't affect the MCP interface.

LGTM! 🚀

PR #85, opened 2026-06-19
fix: produce Altium-loadable PcbLib/SchLib files

ande2407:
## Problem

Libraries written by the tool fail to open in Altium Designer:

- **`.PcbLib`** — `The following exception occurred whilst loading section Library: Catastrophic failure` (and pads display as 0 size)
- **`.SchLib`** — load error on open

## Root causes

Found by diffing generated output byte-for-byte against a real Altium-authored library (`scripts/sample.PcbLib` / `scripts/sample.SchLib`) with `olefile`. The reader shared the writer's offset assumptions, so round-trips looked fine while the files were structurally invalid for Altium.

**PcbLib**
1. **Pad geometry block was 112 bytes** — Altium requires the full **194-byte** pad structure. The truncated block desyncs the parser, which both throws and leaves pad sizes reading as 0.
2. **`/Library` had only `Header` + `Data`** — real libraries also contain `LayerKindMapping`, `PadViaLibrary`, `Textures`, `ModelsNoEmbed`, `EmbeddedFonts`, `ComponentParamsTOC`, and `Models/Header`.
3. **`/Library/Data` carried a minimal synthesized layer stack** that Altium rejects at "section Library".
4. **No per-pad `UniqueIDPrimitiveInformation`** records (real libraries store one UID per pad).
5. **`FileHeader` truncated to 32 bytes** — missing the 8-byte weight and the length-prefixed library UniqueID (real is 53 bytes).

**SchLib**
6. **Binary pin records were 30 bytes vs the required 33** — missing three trailing empty length-prefixed strings after the designator; the reserved byte at offset 13 must be `0x01`.
7. **Missing `Storage` stream** (`|HEADER=Icon storage`).

## Fix

- PcbLib pad geometry emitted from a verified 194-byte template (`PAD_GEOMETRY_TEMPLATE`), with variable fields patched in.
- Added the auxiliary `/Library` substreams in `write_library_extras`.
- `/Library/Data` now splices the caller's `FILENAME` onto a full board/layer-stack config captured from a known-good library (`library_data_template.txt`).
- Always write per-pad `UniqueIDPrimitiveInformation` (matching real libraries, which list pads only).
- FileHeader now includes the weight + library UID.
- SchLib pin records get the 3 trailing string fields and `0x01` reserved byte; `Storage` stream is written.

## Verification

- Regenerated a 4-footprint resistor library (`R0402/0603/0805/2512`) + symbol; both `.PcbLib` and `.SchLib` now open cleanly in Altium Designer with correct pad sizes and linked footprints.
- `cargo test` (207 pass), `cargo clippy`, and `cargo fmt --check` all clean. Updated two unique-id tests for the always-emit-per-pad behaviour.

## Note for maintainers

`/Library/Data` is currently emitted from a verbatim ~95 KB config template captured from `sample.PcbLib`. This is the pragmatic fix that loads reliably; a fully *generated* (non-templated) layer stack would be cleaner long-term. I couldn't bisect the minimal required subset without many Altium round-trips — happy to iterate if you'd prefer that direction.


MatejGomboc:
Thanks for this, @ande2407 — your parallel attempt was genuinely useful. A byte-level audit against the AltiumSharp reference confirmed your core contribution (replacing the synthesised layer stack with a full real-Altium `Library/Data` stack) was the right call, and it's been incorporated into #88 as the `src/altium/pcblib/assets/library_data_stack.txt` asset.

The same audit flagged two byte-level divergences on this branch that we did **not** adopt:

- the pad geometry block here is **194 bytes** (the *PcbDoc* size); the correct *PcbLib* default is **202 bytes** (coupled to byte @172 = 0x1A);
- the SchLib pin `FormalType` reserved byte (0x01) belongs **after** the description Pascal string, not before it — placing it before desyncs every later pin field.

#88 now covers the loadability fix plus a full primitive byte-audit (track/arc extended tails, the canonical 252-byte text record, Windows-1252 string encoding, 24-bit record lengths, and OLE component-name sanitisation), so I'm closing this as superseded. Really appreciate the contribution — it directly improved the fix. 🙏

Issue #68, opened 2026-04-15
[BUG] create pcblib, schlib... altium doesnt read that files


tataremka:
## Bug Description

I just prompt to create:
1. simple footprint 0402
2. simple schlib with resistor.
 nothing less, nathing more

I got files pcblib 6kB, schlib 4kB

Generated files:

[generated libs.zip](https://github.com/user-attachments/files/26762761/generated.libs.zip)

P.S. I have altium 24

achecaquside:
Same issue here, last version.

Any info on how to fix??

MatejGomboc:
Status update after a thorough re-investigation.

**The core problem is fixed.** The writer rewrites in #86 / #87 / #88 (plus #89, #92) made our generated `.PcbLib` / `.SchLib` files Altium-readable. We verify this on every CI run with an **independent reader, `pyaltiumlib`** (`tests/integration/test_altium_readability.py`), which currently passes **13/13 with zero parser warnings** — including after all the recent refactoring.

**Additional hardening just landed/queued.** I ran a byte-level audit of our writer against the **AltiumSharp golden binaries + format spec** and found one more real correctness bug: several string fields were written as **UTF-8 instead of Windows-1252**, which would corrupt any component name/description containing non-ASCII characters (e.g. `10µF`, `±5%`, `°C`). Fixed in #112 (with tests proving non-ASCII now round-trips as Windows-1252). Remaining lower-priority findings (a single-block via rewrite; round-trip fidelity for editing existing files) are tracked in #113.

**What I can't do:** `pyaltiumlib` is a strong proxy, but only Altium Designer itself is the final word. Could whoever still has the failing setup (@/the original reporters) please **regenerate a library with the latest `main` and try opening it in Altium 24**? If it opens cleanly we can close this; if not, a fresh copy of the failing file would let me diff it against the golden references.

(Note: there is still no tagged release, so end users have no prebuilt binary — worth cutting one once this is confirmed.)

achecaquside:
I will take a look, and review your updates and fixes, thanks for the effort in solving! 👍

MatejGomboc:
Closing as **verified fixed**. Summary of what landed since this was reported:

[...]

PR #88, opened 2026-06-20
fix(pcblib): full real-Altium layer stack + pin/stream fidelity (#68)

MatejGomboc:
Hardening of the #68 fixes after cross-checking against (a) the only fork with relevant work — `ande2407/fix:altium-loadable-output`, a parallel AI-assisted attempt — and (b) our own real-Altium `scripts/sample.PcbLib`, byte-for-byte.

> **Stacked on #87 → #86 → #83.** Merge in order.

## Adopted / fixed

**PcbLib `/Library/Data` — the big one.** Our synthesized 9-layer stack (~90 keys) is rejected by Altium with *"Catastrophic failure whilst loading section Library"* (the error ande2407 hit). Replaced it with the **full known-good board/layer-stack config (2092 keys)** captured verbatim from `scripts/sample.PcbLib`, spliced with a dynamic `FILENAME`/`DATE`/`TIME`. Our `Library/Data` now has 2097 keys vs the reference's 2073 — full parity.

**Missing library streams:** added `EmbeddedFonts` (u32 count) and an empty `Models` storage that real libraries carry.

**SchLib binary pin record** (now byte-identical to a real Altium pin):
- FormalType byte (offset 13) = `0x01` (was `0x00`)
- PartAndSequence tail = `"|&|"` (was empty)

## Cross-check outcomes (what I did *not* change)
- **FileHeader middle 8 bytes**: confirmed they ARE the `5.01` version double (verified against the reference: `0a d7 a3 70 3d 0a 14 40`). The parallel fix mis-labeled them "weight" — our 53-byte header was already correct.
- **Pad block size**: stays **202 bytes** (PcbLib). The parallel fix used 194, which is the *PcbDoc* size; the reference `PAD_SMD_RECTANGULAR.PcbLib` block is 202.
- **PrimitiveGuids**: stays omitted — AltiumSharp omits it for from-scratch parts and Altium opens those fine.

## Fork audit (your request)
8 forks: 7 are pure clones (0 commits ahead); only `ande2407` had a relevant branch, now reconciled. The one other change across all forks (`bruce-wang-111`'s manual Windows build) is already superseded by our `release.yml`. **Nothing else to adopt.**

## Verification
`cargo build` · `clippy -D warnings` · `fmt` · full suite (206 lib + integration/property/perf/doctests) · Python E2E 25/25 · **Altium oracle 13/13, 0 regressions**. Generated PcbLib grew 6 KB → ~104 KB (the real Altium stack), as expected.


MatejGomboc:
Follow-up commit on this branch: audited **every user-creatable primitive** against the oracle and found one more malformed record — a footprint **region**'s nested parameter block wasn't null-terminated ("Data does not end with 0x00"), now emitted as a proper C-string block. Also broadened the oracle to a rich footprint (pads+tracks+arcs+text+region) + rich symbol (pins+rectangle+line+text), so all primitives are permanently guarded. Oracle 13/13 with the richer coverage.

MatejGomboc:
## Follow-up: full primitive byte-audit + fixes

Ran a multi-agent byte-level audit of this stack against the AltiumSharp reference (golden binaries + writer source) and the real `scripts/sample.PcbLib`. The original loadability fix holds (oracle 13/13), but the audit surfaced several primitive-fidelity bugs the permissive oracle can't catch. Pushed five commits on top:

- `fix(altium)`: **track/arc extended tails** (were truncated after width; now carry the 16/15-byte subpoly/solder-mask/v7-layer/keepout tail every real Altium track/arc has), SchLib **text → RECORD=4** (was RECORD=3 = Symbol, which Altium mis-renders), `IsNotAccesible=T` on line/label/text, per-pin `IndexInSheet`, OLE name `/`→`_` sanitisation (a `A/B` name was failing to save), and a char-boundary panic fix.
- `fix(schlib)`: **24-bit record length** instead of a truncating `u16` (records >64 KiB silently desynced the stream).
- `fix(altium)`: **Windows-1252 string encoding** (we wrote raw UTF-8; `10µF` now emits the single byte 0xB5, not 0xC2 0xB5) — writer and reader paired.
- `fix(pcblib)`: **canonical 252-byte text record** (was an ~80-byte guessed layout with kind/stroke-width/font-id/v7-layer at the wrong offsets).

Each verified: `cargo build`/`clippy -D warnings`/`fmt`, full suite (211 lib + integration/property/stress/perf/doctests), Altium oracle 13/13 with 0 regressions, plus byte-level confirmation in generated files.

Remaining (documented, not yet fixed — lower urgency, each reader-coupled): pad oval→octagon mapping, pad size/shape (Block 5) corner-radius/slot-hole fidelity, and the PcbLib component-Parameters block still UTF-8. The fork #85 has been reconciled and closed (its useful layer-stack idea is already here).

PR #91, opened 2026-06-21
refactor(altium): unify OLE plumbing + public I/O API (unify pt.3)

MatejGomboc:
Final slice of the PcbLib/SchLib unification: shared OLE entry points and a consistent public I/O surface. Pure refactor — **no emitted/decoded byte changes** (oracle 13/13, round-trip + full suite + E2E green).

## What's unified
- **`crate::altium`** gains `create_ole` / `open_ole` (the v3 compound-file create/open both had duplicated), `write_stream` (create-stream + write-all wrapper), and a shared `generate_ole_names(names)` used by both writers.
- **PcbLib** now exposes `pub read(reader)` / `pub write(&mut self, writer)` (was private `read_from`/`write_to` threading a `&Path` only for tracing). Both types now offer the same `new`/`open`/`save`/`read`/`write` surface.
- SchLib `write` + PcbLib `write_meta_storage` (the 7 metadata storages) route through `write_stream`; both readers open via `open_ole`.

## Left divergent (by design)
- `save(&mut self)` on PcbLib (materialises 3D models first) vs `save(&self)` on SchLib — a genuine difference, not unified.
- Per-format FileHeader/record layouts, per-component storage shapes, dispatch loops — the binary-vs-text format differences.
- The ~12 remaining per-stream `create_stream` sites in PcbLib's `write_footprint`/`write_models` were left as-is: routing them through `write_stream` is pure boilerplate reduction with no byte effect, and not worth churning the #68-tuned paths.

## Verification
`cargo build` · `clippy -D` · full suite · E2E 25/25 · **oracle 13/13, 0 regressions**. MCP layer only calls `.open()`/`.save()` — unaffected.

### DIFF
--- a/src/altium/mod.rs
+++ b/src/altium/mod.rs
@@ -137,6 +137,58 @@ pub fn generate_ole_name<S: BuildHasher>(name: &str, used_names: &HashSet<String
     )
 }
 
+/// Generates collision-free OLE storage names for an ordered list of component
+/// names. Shared by both library writers so the truncation/uniquing rules are
+/// identical; the returned names line up positionally with the input.
+pub(crate) fn generate_ole_names<'a, I>(names: I) -> Vec<String>
+where
+    I: IntoIterator<Item = &'a str>,
+{
+    let mut used = HashSet::new();
+    let mut out = Vec::new();
+    for name in names {
+        let ole = generate_ole_name(name, &used);
+        used.insert(ole.clone());
+        out.push(ole);
+    }
+    out
+}
+
+/// Creates an Altium-mandated OLE v3 (512-byte sector) compound file.
+///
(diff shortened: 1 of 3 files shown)


MatejGomboc:
Follow-up commit: per your nudge that *duplications age poorly*, I went ahead and routed **all** remaining `create_stream` + `write_all` sites in PcbLib (write_models, FileHeader, Library Header/Data, EmbeddedFonts, and the per-footprint Header/Parameters/Data/WideStrings/UID streams) through `write_stream`, and the inline C-string param blocks through `framing::write_cstring_param_block`. No raw `create_stream`+`write_all` remains in production code now. Still byte-identical (oracle 13/13).

PR #95, opened 2026-06-21
fix(scripts): normalise analyze_pcblib input path (alert #7, non-breaking)

MatejGomboc:
Potential fix for [https://github.com/embedded-society/altium-designer-mcp/security/code-scanning/7](https://github.com/embedded-society/altium-designer-mcp/security/code-scanning/7)

The safest fix is to canonicalize (`resolve`) the user-provided path and enforce that it stays within an explicit allowed root directory before using it. This prevents absolute-path abuse and `..` traversal tricks.

Best implementation here (without changing core functionality): in `main()` (around lines 546–550), resolve both `script_dir` and the user path, then verify with `relative_to` that the resolved path is inside `script_dir`. If not, print an error and exit. After that, keep the existing existence check and call `analyze_pcblib` with the validated resolved path. This uses only `pathlib` (already imported), so no new imports/dependencies are needed.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/scripts/analyze_pcblib.py
+++ b/scripts/analyze_pcblib.py
@@ -543,7 +543,14 @@ def main():
             print(f"  Or place a sample file at: {default_path}")
             sys.exit(1)
     else:
-        filepath = Path(sys.argv[1])
+        script_dir = Path(__file__).parent.resolve()
+        filepath = Path(sys.argv[1]).resolve()
+        try:
+            filepath.relative_to(script_dir)
+        except ValueError:
+            print(f"Error: Path must be within {script_dir}: {filepath}")
+            sys.exit(1)
+
         if not filepath.exists():
             print(f"Error: File not found: {filepath}")
             sys.exit(1)
--- a/scripts/analyze_pcblib.py
+++ b/scripts/analyze_pcblib.py
@@ -543,14 +543,11 @@ def main():
             print(f"  Or place a sample file at: {default_path}")
             sys.exit(1)
     else:
-        script_dir = Path(__file__).parent.resolve()
+        # Local developer CLI: the user names the .PcbLib to analyse (see
+        # scripts/README.md), so there is no allow-list to enforce here — pinning
+        # the input under scripts/ would break the documented usage. Resolve the
+        # path to normalise any ".." or symlink components before use.
         filepath = Path(sys.argv[1]).resolve()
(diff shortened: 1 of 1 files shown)


MatejGomboc:
Replaced the Autofix's containment check with a non-breaking hardening.

**Why the original wouldn't work:** the Autofix pinned the input under `scripts/` (`filepath.relative_to(script_dir)`). But `analyze_pcblib.py` is a local developer reverse-engineering CLI whose documented usage (scripts/README.md) is `analyze_pcblib.py /path/to/library.PcbLib` — pointing at an arbitrary `.PcbLib` anywhere on disk. Restricting to `scripts/` breaks that. There is no privilege boundary here (this script is not the MCP server — that's the Rust binary in `src/`, which already enforces an allow-list via `validate_path`), so there is no safe root to contain to.

**What this now does:** `Path(sys.argv[1]).resolve()` to normalise `..`/symlink components, then the existing existence check. Usage is preserved.

Note: CodeQL may still flag the `argv → Path → read` dataflow since there's no containment (there can't be one for a "read any file the user names" tool). If it does, alert #7 is best **dismissed as a false positive** (local CLI, no trust boundary) rather than worked around.

MatejGomboc:
Closing without merging.

The change here (`Path(sys.argv[1]).resolve()`) is harmless but a no-op for its stated purpose: with no containment it adds no real security, and it does not silence CodeQL — the argv → Path → read dataflow is still flagged. It cannot be contained, because this is a "read any file you name" local developer CLI (see scripts/README.md). The original Copilot Autofix (pinning the input under `scripts/`) would have broken that documented usage.

The correct resolution for alert #7 is to dismiss it as a **false positive**: `analyze_pcblib.py` is a standalone local dev CLI with no trust boundary — the user runs it from their own shell and names the file to read. The actual MCP-server trust boundary is enforced separately in the Rust binary (`src/mcp/server.rs::validate_path` — fail-closed allow-list with canonicalised `starts_with` containment).

A dismissal request for alert #7 (reason: *false positive*) has been filed and is pending approval via delegated dismissal.

Note: this does not touch the separate open alert #6 (`rust/path-injection`, `src/mcp/server.rs:860`) — that one sits on the real trust boundary and is worth a proper review.

PR #115, opened 2026-06-22
fix(schlib): link footprint models so all resolve in Altium

ande2407:
## Problem

A symbol written with multiple footprint models showed **only the first** in Altium's SCH Library editor, and each footprint's model preview reported **"footprint not found."**

## Root causes

Found by diffing a tool-written symbol against the same symbol after adding a footprint by hand in Altium.

1. **`OwnerIndex`** — `encode_data_stream` passed each model's *enumerate index* (`0,1,2,3`) as its `RECORD=45` `OwnerIndex`. But `OwnerIndex` must reference the single `RECORD=44` ImplementationList that owns every model. Altium attached only the model whose owner happened to resolve (the first) and orphaned the rest.
2. **Resolution** — models were written `DatafileCount=0` (name only), so Altium couldn't bind them to a footprint or render a preview. Real models carry `DatafileCount=1` + `ModelDatafileEntity0` + `ModelDatafileKind0`, and a `ModelDatafile0` path.

## Fix

- Compute the `RECORD=44` stream-index (new `count_records` helper) and use it as `OwnerIndex` for **all** models; emit each model's `RECORD=46` + `RECORD=48` child records owned by their `RECORD=45`.
- Write `DatafileCount=1` + entity/kind + `IsCurrent`; add an optional `FootprintModel.library_path` → `ModelDatafile0`, plumbed through `manage_schlib_footprints` (new optional `library_path` arg) and `write_schlib` (per-footprint `library_path`), and parsed back by the reader so it round-trips.

## Verification

- A resistor symbol with `R0402`/`R0603`/`R0805`/`R2512` now shows **all four**, and the model previews **render** (was "not found"). Confirmed in Altium Designer.
- `cargo test` (225, incl. a new regression test that asserts every model is owned by the `RECORD=44` and the path is emitted), `cargo clippy`, and `cargo fmt --check` are clean.

## Notes for maintainers

- `library_path` is the one small **API addition** (optional arg on `manage_schlib_footprints`, and per-footprint in `write_schlib`). When omitted, models link by name only — which works when the PcbLib is installed/in the project, but reports "not found" otherwise. Happy to rename it or auto-derive the path if you'd prefer a different shape.

### DIFF
--- a/src/altium/schlib/primitives.rs
+++ b/src/altium/schlib/primitives.rs
@@ -896,6 +896,13 @@ pub struct FootprintModel {
     /// Description.
     #[serde(default, skip_serializing_if = "String::is_empty")]
     pub description: String,
+    /// Path to the `PcbLib` that contains this footprint, written as
+    /// `ModelDatafile0`. When set, Altium resolves the footprint directly from
+    /// that file (rendering the preview); when absent it falls back to searching
+    /// available libraries by name, which reports "footprint not found" if the
+    /// library isn't installed/in the project.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub library_path: Option<String>,
 }
 
 impl FootprintModel {
@@ -905,6 +912,7 @@ impl FootprintModel {
         Self {
             name: name.into(),
             description: String::new(),
+            library_path: None,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


MatejGomboc:
Thanks for this, @ande2407! 🙏 Merged. Linking the footprint models through the implementation list so they all resolve in Altium is a great fix. (I resolved a small conflict with #119 in `encode_data_stream`'s tests and ran rustfmt — both changes are in.)

PR #116, opened 2026-06-22
feat(libpkg): generate Altium Library Package (.LibPkg) project files

ande2407:
## What

Adds a `write_libpkg` MCP tool that generates an Altium **Library Package** (`.LibPkg`) — the project file that groups a `.SchLib` + `.PcbLib` so Altium can **Compile Integrated Library** into a single `.IntLib`.

## Why

The tool could already write source `.SchLib`/`.PcbLib` files but had no way to package them into a combined library. `.LibPkg` is the standard source for an Integrated Library, so this closes the gap from "two source files" to "one compilable library project."

## Format

A `.LibPkg` is an INI-style project file. Real ones run ~1000 lines, but Altium backfills every optional setting on open — only this is required, with members referenced by path **relative to the `.LibPkg`**:

```ini
[Design]
Version=1.0

[Document1]
DocumentPath=MyLib.SchLib

[Document2]
DocumentPath=MyLib.PcbLib
```

## Implementation

- **`altium::libpkg` module** — `build_libpkg` + lexical relative-path computation (same-folder → bare name; subdirectory and parent handled; bare name / different-drive passed through). 5 unit tests.
- **`write_libpkg` tool** — `filepath` + an explicit `documents` list, wired through dispatch and the tool schema.
- **Scope boundary** — generates project *source* only; compiling to the binary `.IntLib` is a one-click step in Altium (we can't produce the compiled binary). Documented in the tool description and module docs.

## Verification

- Generated a `.LibPkg` for a resistor `.SchLib` + `.PcbLib` pair; Altium opened it with both members and **Compile Integrated Library** produced a clean `.IntLib`.
- `cargo test` (229, +5 new), `cargo clippy`, `cargo fmt --check` all clean.

## Notes for maintainers

- The document list is **explicit** rather than auto-discovered, for predictability (no sweeping in unrelated sibling libraries). An auto-discover convenience flag could be added later if desired.

### DIFF
--- a/src/altium/libpkg.rs
+++ b/src/altium/libpkg.rs
@@ -1,0 +1,123 @@
+//! Altium Library Package (`.LibPkg`) project-file generation.
+//!
+//! A `.LibPkg` is an INI-style project file that groups source library
+//! documents (`.SchLib`, `.PcbLib`) so Altium can compile them into an
+//! Integrated Library (`.IntLib`). Altium backfills every optional project
+//! setting the first time the project is opened, so only the `[Design]`
+//! version and one `[Document{N}]` section per member document are required.
+//!
+//! ```text
+//! [Design]
+//! Version=1.0
+//!
+//! [Document1]
+//! DocumentPath=MyLib.SchLib
+//!
+//! [Document2]
+//! DocumentPath=MyLib.PcbLib
+//! ```
+//!
+//! Member documents are referenced by their path **relative to the `.LibPkg`
+//! file**. This module only generates the project source; compiling it to a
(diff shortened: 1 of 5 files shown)


MatejGomboc:
Thanks for this, @ande2407! 🙏 Merged. Really useful feature — generating the `.LibPkg` project so libraries can compile to an IntLib. One small cross-platform follow-up: the relative-path helper used `std::path::Path::components()`, which only treats `\` as a separator on Windows — so the unit tests passed on Windows but the Build job failed on Linux/macOS. Fixed in #120 (we split the Windows-style paths ourselves); behaviour on Windows is unchanged.

PR #119, opened 2026-06-22
fix(schlib): write body rectangle before pins so the fill doesn't hide pin names

ande2407:
## Problem

A schematic symbol with a solid-filled body (a filled rectangle — the normal case) loses its **pin names**: they render and are then immediately painted over by the body fill, so the symbol shows pins and numbers but no names.

## Cause

`encode_data_stream` writes records in the order *component → parameters → **pins → rectangles***. Schematic primitives are drawn in record order, so the filled rectangle (written after the pins) draws on top and covers the pin names that sit inside the body. Pin names *outside* the body are unaffected — which is why the bug only appears once the pins sit flush against a filled body.

## Fix

Write rectangles **before** pins so the body sits at the back — matching Altium's own ordering (in real `.SchLib` files the body rectangle precedes the pins). One-block reorder in `encode_data_stream`, plus a regression test that asserts the rectangle record precedes the first pin record.

## Verification

- Built a real IC symbol (Silergy SY8493, SO8E, solid yellow body, 9 pins): pin names now render correctly on top of the fill.
- `cargo test` (225, incl. the new `body_rectangle_is_written_before_pins`), `cargo clippy`, `cargo fmt --check` all clean.

### DIFF
--- a/src/altium/schlib/writer.rs
+++ b/src/altium/schlib/writer.rs
@@ -602,18 +602,21 @@ pub fn encode_data_stream(symbol: &Symbol) -> crate::altium::error::AltiumResult
         index_counter += 1;
     }
 
-    // 3. Pins (binary format)
+    // 3. Rectangles — written before the pins so the body shape sits at the
+    //    back. Emitting pins first lets a solid-filled body paint over the pin
+    //    names that sit inside it (names vanish). This matches Altium's own
+    //    ordering (body rectangle precedes pins in its symbol records).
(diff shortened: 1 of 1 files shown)


MatejGomboc:
Thanks for this, @ande2407! 🙏 Merged. Nice catch on the draw order — emitting the body rectangle before the pins so a solid fill doesn't paint over the pin names is exactly right, and it matches Altium's own record ordering.

PR #124, opened 2026-06-23
Fix Region layer: emit canonical V7_LAYER token (MECHANICAL{n})

ande2407:
## Problem

A footprint Region placed on a component-pair / mechanical layer (e.g. `Top Courtyard`) renders on **Top Layer (copper)** in Altium and does **not** appear in the footprint primitives list.

Repro: write a footprint with a `regions` entry on `"Top Courtyard"`, open the `.PcbLib` in Altium → giant copper region covering the pads, absent from the Footprint Primitives panel.

## Root cause

`encode_region_properties` built the `V7_LAYER` param from the display name with spaces stripped:

```rust
let layer_name = region.layer.as_str().replace(' ', "").to_uppercase(); // "TOPCOURTYARD"
```

Altium keys a Region's layer off the `V7_LAYER` **string param**, not the common-header layer byte. `TOPCOURTYARD` is not a valid Altium layer token (the layer's canonical token is `MECHANICAL4`), so Altium fails to resolve it and falls back to Top Layer.

The common-header layer byte was written correctly (`60` = Mech 4), so the MCP reader (which reads the byte) round-tripped the region fine — masking the bug. It only shows up in Altium itself.

This affects every region on a named component-layer alias (`Top Courtyard`, `Top Assembly`, the 3D-body layers, etc.). It works for `Top Layer`/`Top Overlay` only because their stripped display name happens to equal the canonical token. The 3D-body encoder already sidesteps this by hardcoding `MECHANICAL6`/`MECHANICAL7`.

## Fix

Map the layer to its canonical `V7_LAYER` token — `MECHANICAL{n}` for mechanical/component-pair layer IDs (57–72, 186–201), stripped display name otherwise — and use it in `encode_region_properties`.

Verified end-to-end: regenerating a WSON-8 footprint with a `Top Courtyard` region now writes `V7_LAYER=MECHANICAL4`; Altium places it on Top Courtyard and lists it as a Region.

Added a unit test for the token mapping and the serialized properties block.

### DIFF
--- a/src/altium/pcblib/writer.rs
+++ b/src/altium/pcblib/writer.rs
@@ -896,6 +896,22 @@ fn encode_region(data: &mut Vec<u8>, region: &Region) {
     write_block(data, &[]);
 }
 
+/// Returns the canonical Altium `V7_LAYER` token for a layer.
+///
+/// Altium identifies a Region's layer by this parameter string, NOT the
+/// common-header layer byte. Mechanical and component-pair layers must use their
+/// `MECHANICAL{n}` token (e.g. Top Courtyard -> `MECHANICAL4`). The display name
+/// with spaces stripped (`TOPCOURTYARD`) is not a valid token, so Altium fails to
+/// resolve the layer and silently drops the region onto Top Layer (copper). The
+/// 3D-body encoder already hardcodes `MECHANICAL6`/`MECHANICAL7` for this reason.
+fn region_v7_layer_token(layer: Layer) -> String {
+    match layer_to_id(layer) {
+        id @ 57..=72 => format!("MECHANICAL{}", id - 56),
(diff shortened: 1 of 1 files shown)


MatejGomboc:
Thanks for this, @ande2407! 🙏 Merged. Excellent catch — and exactly the kind of bug only real Altium can reveal: the region round-tripped fine through our reader and the `pyaltiumlib` oracle because both read the common-header **layer byte**, while Altium keys the region's layer off the `V7_LAYER` **token**. Mapping mechanical/component-pair layers to their canonical `MECHANICAL{n}` token (with the unit test) is the right fix. Validated locally against `main` — build, clippy, full tests, and the readability oracle all green.

PR #133, opened 2026-06-23
Add generic extruded 3D bodies to write_pcblib

ande2407:
## Summary

Adds support for generic **extruded 3D bodies** to `write_pcblib`. Callers can now attach a `ComponentBody` to a footprint from an outline polygon plus standoff/overall height — no external STEP model required — and have it render in Altium's 3D view. A ceramic-capacitor library (01005 → 2512, each with a courtyard region, assembly outline, designator, and an extruded body) was the driving sample and loads cleanly with all bodies rendering.

## What's included

**Extruded ComponentBody encoding** (`build_component_body_params` / `encode_component_body`)
- Generic extruded body when no model name is supplied: `ISSHAPEBASED=FALSE`, `MODEL.MODELTYPE=0`, `MODEL.EXTRUDED.MINZ/MAXZ` from standoff/overall height, and a synthesized `MODELID` GUID. Model-backed (STEP) bodies keep `MODELTYPE=1`.
- Outline vertices written as **whole Altium internal units** (via `from_mm`), like every other primitive. Fractional internal coordinates caused Altium to silently discard the body.
- `V7_LAYER` derived from the body's actual layer (`region_v7_layer_token`) rather than a hardcoded `MECHANICAL6`; a token/layer-byte mismatch made Altium drop the body.

**Region single-block fix** (writer + reader)
- A Region serializes as a **single** length-prefixed block. The encoder was appending a second, empty block; Altium reads the region's one block, lands on the stray `00 00 00 00`, treats it as an invalid record type, and **silently drops every primitive after the region**. With a `ComponentBody` following the courtyard region, the body never rendered. Verified against Altium-resaved libraries that the next record's type byte follows the region's block immediately.
- The reader no longer consumes a phantom second block, keeping round-trips consistent and correctly reading Altium-authored region→body sequences.

**MCP surface**
- `component_bodies` input on `write_pcblib` (schema + parsing): `outline`, `overall_height`, `standoff_height`, `layer`, rotations.

## Testing
- New regression test: `encode_region` emits exactly one block (guards the dropped-primitive bug, which round-trip tests miss since reader+writer were self-consistent).
- Extruded-vs-model parameter coverage.
- Full lib suite green (236 tests); clippy clean.
- **Manually verified in Altium**: bisection isolated the region as the culprit, and after the fix every footprint in the capacitor library shows its extruded body in the 3D view.


MatejGomboc (approved):

MatejGomboc:
Merged — thank you, @ande2407! 🙏 Beautiful, Altium-verified work. The extruded-body encoding is spot on: deriving `V7_LAYER` from the real layer (instead of a hardcoded `MECHANICAL6`), `MODELTYPE=0` for extruded vs `1` for model-backed, `MODELSOURCE` only when non-extruded, and whole-unit outline coords — every one of those matches what an independent byte-level RE of the format turned up, so it's great to see them confirmed in real Altium.

Heads-up on the merge: your region single-block fix landed in parallel with #132, which fixed the same bug, so I reconciled the overlap (kept one copy, combined both sets of comments, and preserved your `encode_region_emits_exactly_one_block` test — nice catch that round-trip tests miss it). I also tidied a `cargo fmt` diff and a clippy doc-markdown lint so the gated CI went green. All verified locally too: clippy clean, full suite + the readability oracle 13/13.

Really appreciate the careful bisection-in-Altium writeup. 🍻

PR #153, opened 2026-06-24
feat(designators): always assign SchLib designator + auto-inject PcbLib .Designator

ande2407:
## Summary

Make reference designators reliable in both library types.

- **PcbLib — always include the `.Designator` directive.** Every footprint now gets a `.Designator` special string on the Top Overlay when the caller did not supply one. This is intentionally **independent of the SchLib**: even if a symbol leaves its designator blank, the footprint must still carry the `.Designator` placeholder so a placed component renders its reference designator on the board. An existing `.Designator` text is left untouched (case-insensitive check).
- **SchLib — always assign a designator.** Previously a symbol written without `designator_prefix` got *no* designator record at all. We believe it is in the best interest to always include a designator, so the tool now decides one from the information provided:
  1. explicit `designator_prefix` (unchanged)
  2. new optional `component_type`, mapped via an IEEE 315 / ASME Y14.44 class-letter table (R, C, L, D, Q, U, J, Y, …)
  3. `"U"` fallback (IC / inseparable assembly — the most common case)

## Details

- `.Designator` placement: `x = 0`, just above the topmost pad (origin if no pads), 1 mm stroke, Top Overlay; the user can reposition in Altium.
- `component_type` is documented in the `write_schlib` schema; unknown values resolve to `U`.
- Unit tests cover the IEEE mapping (known types, case/whitespace insensitivity, unknown → `U`).

## Testing

- `cargo test` green (incl. new unit tests).
- `cargo clippy --all-targets --all-features -- -D warnings` clean.
- Validated end-to-end by regenerating a real 19-symbol / 9-footprint TI INA2xx library: every footprint now carries `.Designator`, every symbol carries `U?`.

## Notes / open questions

- Diodes map to `D` (commercial convention), not the mil-spec `CR`. Happy to make the diode/zener split configurable if the maintainers prefer.
- PcbLib injection is Top Overlay only (top-side authoring).

### DIFF
--- a/src/mcp/tool_definitions.rs
+++ b/src/mcp/tool_definitions.rs
@@ -340,7 +340,8 @@ impl McpServer {
                                 "properties": {
                                     "name": { "type": "string" },
                                     "description": { "type": "string" },
-                                    "designator_prefix": { "type": "string", "description": "e.g., 'R' for resistors, 'U' for ICs" },
+                                    "designator_prefix": { "type": "string", "description": "Reference-designator class letter, e.g. 'R' for resistors, 'U' for ICs. Written as '<prefix>?'. If omitted, falls back to 'component_type' (IEEE 315 / ASME Y14.44 mapping), then to 'U'." },
+                                    "component_type": { "type": "string", "description": "Optional component category (e.g. 'resistor', 'capacitor', 'inductor', 'diode', 'transistor', 'connector', 'crystal', 'ic') used to derive the IEEE designator letter when 'designator_prefix' is not given. Unknown values default to 'U'." },
                                     "part_count": { "type": "integer", "description": "Number of parts for multi-part symbols (e.g., 2 for dual op-amp). Default: 1" },
                                     "pins": {
                                         "type": "array",
--- a/src/mcp/tools/read_write.rs
+++ b/src/mcp/tools/read_write.rs
@@ -4,6 +4,51 @@ use serde_json::{json, Value};
 
 use crate::mcp::server::{ErrorContext, McpServer, ToolCallResult};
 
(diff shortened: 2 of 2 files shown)


MatejGomboc:
Thank you for this, @ande2407 — genuinely lovely work, and a real quality-of-life win. 🙌 Always-present designators in both library types is exactly the kind of thing that makes generated libraries feel finished, and the IEEE-315 / ASME-Y14.44 class-letter table is comprehensive and correct. The unit tests (known types, case/whitespace insensitivity, unknown → `U`) and the real 19-symbol TI-library validation are appreciated too.

**Heads-up: I rebased your branch onto `main` and pushed one small fixup.** A change that landed in the meantime (#154) added a new `Text.italic` field, so your auto-injected `.Designator` `Text { … }` literal no longer compiled. I added `italic: false` (the stroke-text default — byte-identical) in a clearly-labelled commit on top of yours; your work is untouched otherwise.

I then ran the full gate and everything is green: `cargo build`, `cargo test`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt --all --check`, **and the pyaltiumlib readability oracle stays 13/13** — which confirms the injected `.Designator` is valid Altium output (the oracle exercises the `write_pcblib` path your inject lives on).

On your open questions: diode → `D` (commercial) is the right default for this tool; no need to split out mil-spec `CR` unless someone asks. Top-Overlay-only injection is fine for top-side authoring. Both are reasonable as-is.

CI shows as blocked only because it's an external-fork PR — that's just GitHub waiting on a maintainer to approve the workflow run, not anything in your code. Merging shortly. Thanks again for contributing! 🇬🇧

PR #161, opened 2026-06-24
fix(schlib): correct default fill colour from cyan to Altium light yellow

ande2407:
## Summary

The default schematic **fill colour was wrong** — `0xFFFFB0`. In Altium's BGR packing that is light **cyan** (RGB 176,255,255); it's the intended light-yellow value with the blue/red bytes swapped. So any filled `Rectangle` / `Ellipse` / `RoundRect` authored without an explicit `fill_color` — and any rectangle/ellipse written via `write_schlib` without one — rendered **cyan** in Altium instead of the expected pale yellow.

This corrects the default to **`0xB0FFFF`** (= `11599871`, RGB 255,255,176), which is the exact value the component header already emits as `AreaColor` in `schlib/writer.rs`. Primitives now match Altium's own default schematic fill, and the header/primitive values are internally consistent.

## Changes

- `schlib/primitives.rs` — `Rectangle::new`, `Ellipse::new`, `RoundRect::new` defaults `0xFFFFB0` → `0xB0FFFF`.
- `mcp/tools/parsing.rs` — both `write_schlib` `fill_color` parse fallbacks corrected.
- `mcp/tool_definitions.rs` — schema doc updated to `Default: 0xB0FFFF`.
- Regression test asserting the three constructors default to `11599871`.

## Testing

- `cargo test` green (incl. new `filled_shapes_default_to_altium_light_yellow`).
- `cargo clippy --all-targets --all-features -- -D warnings` clean.

## Note

This is a straight value correction (cyan → the already-used yellow). If you'd rather the default fill be **configurable** (e.g. a config setting, or always caller-specified) instead of hardcoded, happy to rework — but this keeps current behaviour and just fixes the colour to Altium's actual default.


MatejGomboc (dismissed):

MatejGomboc:
Lovely fix — thank you! The diagnosis is exactly right: `0xFFFFB0` in Altium's COLORREF/BGR packing is RGB(176,255,255) = light **cyan**, and the `// Light yellow` comment confirms yellow was always the intent. The clincher is internal consistency — `schlib/writer.rs` already emits `AreaColor=11599871` (= `0xB0FFFF`) in the component header, so the primitive defaults were out of step with the library's own header. 

I verified locally on your branch: `cargo fmt`/`clippy` clean, full suite green (incl. your new `filled_shapes_default_to_altium_light_yellow`), and the independent `pyaltiumlib` readability oracle stays 13/13.

One small thing the PR missed: `docs/SCHLIB_FORMAT.md` still listed the old `0xFFFFB0` as "Light yellow" — perpetuating the very swap you fixed. Since you'd kindly enabled *maintainer edits*, I pushed a one-line commit to your branch correcting that table cell, so the docs match the code. Hope that's alright!

Merging shortly. Thanks again for the sharp contribution. 🙏

PR #182, opened 2026-06-25
feat(pcblib): auto-add extruded 3D body + actionable height echo

ande2407:
## Summary

Two related changes to `write_pcblib`:
1. **Auto-add an extruded 3D body** to any footprint written with no STEP model and no `component_bodies` (and at least one pad). The outline is synthesised from the pads; height defaults to **1.0 mm**, sitting flush (standoff 0).
2. **Echo each footprint's 3D body** back in a new `bodies` array — `source` (`step-embedded` / `step-external` / `extruded` / `auto-extruded` / `none`), `overall_height`, `standoff_height`, and `assumed_height`. When the height was the auto default, the entry also carries an **`action_required`** message telling the caller it's a placeholder and to look up the real height and re-write.

## Motivation — first-run quality of life

First impressions decide whether a tool gets a second chance. When someone points a *fresh* agent at this MCP for the first time, the agent has none of their conventions or feedback yet — and that's exactly when output tends to be rough. A footprint that silently ships with **no 3D body**, or a wrong one, is the kind of early papercut that makes a new user conclude "it doesn't work" and walk away before the agent has had any chance to learn what they want.

This is a quality-of-life safety net for that initial window:
- Every footprint gets a sensible 3D body by default, so a placed part is never flat/empty in 3D.
- Any guessed value is made **loud and self-correcting** rather than silent — the response tells the agent the height is a placeholder and to replace it.

The goal is to turn the most common first-run failure mode into a self-fixing nudge, so people don't abandon the tool over a first-try miss while it's still learning their preferences.

## In practice

Capable agents usually research and set the real height up front, so the 1.0 mm default never persists for them. When an agent *does* omit it, the `action_required` message reliably drives self-correction — verified end-to-end: a fresh agent that wrote a body-less inductor footprint received the message, pulled the datasheet height, and rewrote with `overall_height: 6.2` (`assumed_height` → false). Either way, the placeholder shouldn't survive into a finished part.

## Testing

- Unit test: body summary reports the right source/height, and `action_required` is present only in the assumed case.
- End-to-end: auto-body fires with a pad-derived outline; the assumed-height self-correction loop completes.
- `cargo test` green; `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo fmt` clean.

### DIFF
--- a/src/mcp/tool_definitions.rs
+++ b/src/mcp/tool_definitions.rs
@@ -149,8 +149,13 @@ impl McpServer {
                      regions, and text. The AI is responsible for calculating correct positions \
                      and sizes based on IPC-7351B or other standards. \
                      All coordinates and dimensions must be in millimetres (mm). \
-                     The response includes a 'warnings' array flagging silkscreen (overlay) \
-                     tracks that overlap a pad (silk-on-pad) so you can move them clear."
+                     If a footprint has no STEP model and no component body, an extruded 3D body \
+                     is added automatically (default height 1.0 mm). The response 'bodies' array \
+                     echoes each footprint's 3D body height and source ('auto-extruded' bodies are \
+                     flagged 'assumed_height': true) so you can confirm or override the height by \
+                     supplying 'component_bodies' explicitly. The response also includes a \
+                     'warnings' array flagging silkscreen (overlay) tracks that overlap a pad \
+                     (silk-on-pad) so you can move them clear."
                         .to_string(),
                 ),
                 input_schema: json!({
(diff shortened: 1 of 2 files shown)


ande2407:
When I first started on Claude, I ran into this. When I started testing in Gemini, I ran into this. Once you tell your agent to fix it, it never occurs again, but I thought it might be better to just have the tool echo back to the agent that it missed something. The agents seem to respond well to this corrective action seamlessly behind the scenes. I had Claude try and spawn fresh agents, but they also didn't run into it. So this might never get used in practice. I like the precedent of the tool helping teach the agent though.

MatejGomboc (approved):

PR #203, opened 2026-06-30
feat(mcp): self-documenting tools — generated TOOLS.md, drift guard, initialize instructions

ande2407:
## Description

Agents that consume a **release** have no source tree, and the MCP surface today is tools-only — so the conventions that aren't per-tool (coordinate units, the pin body-attach/orientation rule, the path sandbox, the build flow) lived only in `docs/`/source, and agents reached straight for the source to answer API questions. This adds two **self-updating** channels, both sourced from code so neither can drift, plus tests that enforce that.

### 1. `docs/TOOLS.md` is now generated from the tool definitions
Rendered from `get_tool_definitions()` — the same defs served over `tools/list` — so the human-readable reference can't disagree with the schema. Each `ToolDefinition` carries an optional `example` (`#[serde(skip)]`, internal-only) rendered with its description and a parameter table derived from the input schema.

- Regenerate: `UPDATE_DOCS=1 cargo test --lib tools_md_in_sync`
- A unit test (`tools_md_in_sync`) fails if the committed doc is stale — same guarantee `tools/list` has. This also removes the drift risk in the previously hand-maintained `TOOLS.md`.

### 2. `initialize` now sends MCP `instructions`
Embedded via `include_str!` from a new `docs/AGENT_GUIDE.md` (invariants only — it points at `tools/list` for per-tool detail, so adding a tool needs no edit there). Clients surface this to the model automatically, which is the protocol's designated "how to use me" channel.

### 3. Release bundles the agent docs
`release.yml` now ships `docs/AGENT_GUIDE.md` + `docs/TOOLS.md` in `dist/docs/`, so a downloaded release teaches the consuming agent without the source tree.

## Type of Change

- [x] CI/CD changes (release artifact contents)

## Testing

- [x] `tools_md_in_sync` — generated doc vs committed `TOOLS.md`
- [x] `examples_are_schema_valid` — every example names its tool, uses only documented args, supplies required args. **Caught two stale examples** harvested from the old hand-written doc: `restore_backup` used `backup_filename` (real param `backup_path`); `bulk_rename` referenced a nonexistent `pattern_type`. Both fixed.
- [x] `example_field_is_not_serialized` — the internal `example` stays off the `tools/list` wire
- [x] `initialize_sends_agent_instructions` — `initialize` returns the embedded `AGENT_GUIDE.md`
- [x] `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test` (319 unit) green
- [x] Integration `tests/integration/test_mcp_tools.py` 131/131; `ORACLE_REQUIRE_DEPS=1 .../test_altium_readability.py` 0 regressions

## Notes

The `example` data is compiled into the binary but `#[serde(skip)]` and only read by the (test-only) doc generator — it never reaches `tools/list` and has no runtime effect. The `release.yml` doc-copy only exercises on a tag build; the referenced paths were verified to exist.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -148,6 +148,8 @@ jobs:
                 cp target/${{ matrix.target }}/release/${{ matrix.binary }} dist/
                 cp README.md LICENCE CHANGELOG.md dist/
                 cp config/example-config.json dist/
+                mkdir -p dist/docs
+                cp docs/AGENT_GUIDE.md docs/TOOLS.md dist/docs/
                 cd dist
                 tar -czvf ../${{ matrix.artifact }}.tar.gz *
 
@@ -159,6 +161,8 @@ jobs:
                 cp target/${{ matrix.target }}/release/${{ matrix.binary }} dist/
                 cp README.md LICENCE CHANGELOG.md dist/
                 cp config/example-config.json dist/
+                mkdir -p dist/docs
+                cp docs/AGENT_GUIDE.md docs/TOOLS.md dist/docs/
                 cd dist
                 7z a -tzip ../${{ matrix.artifact }}.zip *
 
--- a/docs/AGENT_GUIDE.md
+++ b/docs/AGENT_GUIDE.md
@@ -1,0 +1,55 @@
+# Using altium-designer-mcp
+
+Invariants for an agent driving this server. For the per-tool parameter schema
+and a worked example of every tool, call `tools/list` (the schema travels with
+the server) or read `docs/TOOLS.md`. This guide covers only the conventions that
+the per-tool schema does not.
+
+## Responsibility split
+
+The **agent** supplies the intelligence — it computes footprint land patterns,
+pin positions, and symbol geometry (from datasheets, IPC-7351, etc.). The
+**tool** only validates and writes those primitives to the Altium binary
+format. The server never invents geometry; pass it exact, finished numbers.
+
+## Coordinate units (this trips up most callers)
+
+- **`.PcbLib` (footprints):** millimetres. Pad/track/arc/region coordinates and
+  sizes are all mm.
+- **`.SchLib` (symbols):** schematic units where **10 units = 1 grid square**
+  (≈ 100 mil). These are *not* millimetres. Graphic primitives accept fractional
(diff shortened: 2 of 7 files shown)


MatejGomboc (approved):

MatejGomboc (approved):
Reviewed in depth (isolated worktree) — correct and well-built. Verified: the internal `example` (#[serde(skip)]) never reaches `tools/list`; every `AGENT_GUIDE.md` invariant cross-checks against the code (mm units, the ~100-mil SchLib grid, the pin body-attach/tip-direction rule vs pin_tip(), the allowed_paths sandbox + timestamped .bak, the build flow); the drift guard (tools_md_in_sync) + examples_are_schema_valid work (it caught two stale examples); release.yml doc-copy is correct and tag-gated; CHANGELOG/CoC untouched; Rust gates all green (fmt/clippy/cargo doc -Dwarnings/cargo test 319). The one integration hiccup — cp1252 decode of the new non-ASCII `initialize` instructions — was a latent harness bug, fixed independently in #206; with it, integration 131/131 + oracle 0 regressions. A genuinely useful self-updating docs channel. Thanks @ande2407!

MatejGomboc:
Thanks so much for this, @ande2407 — merged! 🎉

The self-documenting-tools idea is genuinely valuable: a generated `TOOLS.md` that can't drift from the schema, per-tool examples, and `initialize` shipping the `AGENT_GUIDE.md` instructions are exactly what makes the server usable from a release without the source tree. The in-depth review confirmed it's cleanly built — the internal `example` stays off the `tools/list` wire, the drift guard + schema-valid checks work (nice catch on the two stale examples), and every `AGENT_GUIDE.md` invariant checks out against the code.

Two tiny follow-ups landed on `main` right after — nothing on you:
- **#206** fixed a latent test-harness bug (it decoded the server's UTF-8 stdout with the OS locale codec) that your non-ASCII `initialize` text was simply the first to surface.
- **#207** added a `markdownlint-disable MD013` to the generated `TOOLS.md`, since its parameter tables and inline JSON schemas legitimately run past the 170-char line limit.

Really appreciate the contribution — thanks again! 🙏

PR #294, opened 2026-08-10
fix(mcp): scope write_schlib geometry echo to the symbols written

ande2407:
## Description

`write_schlib`'s `geometry` echo built its array from `library.iter()` — the **whole** library — rather than the symbols the call actually wrote. With `append: true` that makes the response grow quadratically with the number of appends.

Hit this while batch-building a 27-symbol library over 11 append calls (library counts 4 → 9 → 11 → … → 27):

- symbol-geometry blocks echoed: **196**
- blocks actually needed: **26**

With 105-pin symbols in the set, the later responses were large enough to stop being usable. Beyond size, echoing pre-existing symbols defeats the echo's stated purpose — the comment above it says it exists "so the caller can verify pin placement", i.e. verify *this* write.

## Fix

Record the names added during the write loop and filter the echo to that set.

Names are collected in the loop rather than reused from the existing `new_names` vector, because `new_names` is built with `filter_map` on the `"name"` field — a symbol that omits `name` is written under the default but would be missing from the echo.

`symbol_names` still reports the full library: that is intended (it describes the resulting library) and is cheap.

`write_pcblib` is **unaffected** — its `bodies` echo is built inside the footprint loop and was already scoped to the footprints written.

## Where it came from

Introduced in #170, which added the geometry echo. Reported and fixed against current `main`.

## Type of Change

## Testing

- [x] New regression test `write_schlib_geometry_echo_covers_only_written_symbols`. Verified it fails on the pre-fix code:
  ```
  left:  ["RESISTOR", "CAPACITOR", "APPENDED"]
  right: ["APPENDED"]
  ```
- [x] `cargo test` — 716 lib + all integration suites pass (incl. `mcp_e2e` 36/36)

## Notes

Output-shape change only for callers that were reading geometry entries for symbols they did not write in that call — which was not the documented behaviour. No schema or input change.

### DIFF
--- a/src/mcp/server.rs
+++ b/src/mcp/server.rs
@@ -2226,6 +2226,49 @@ mod tests {
         assert!(lib.get("NEW_SYM").is_some());
     }
 
+    #[test]
+    fn write_schlib_geometry_echo_covers_only_written_symbols() {
+        // The geometry echo exists so the caller can verify the pins it just wrote.
+        // It previously walked the whole library, so an `append: true` sequence
+        // re-echoed every pre-existing symbol and the response grew quadratically
+        // with the number of appends. Assert it is scoped to this call's symbols.
+        let temp = test_temp_dir();
+        let lib_path = temp.path().join("geom_scope.SchLib");
+        create_test_schlib(&lib_path); // seeds RESISTOR + CAPACITOR
+
+        let server = create_test_server(temp.path());
+        let args = json!({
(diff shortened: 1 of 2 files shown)


MatejGomboc:
Reviewed — good catch on the quadratic echo growth, and recording written_names at add-time (so default-name fallbacks are covered) is carefully done. Thank you @ande2407!

MatejGomboc (approved):

PR #295, opened 2026-08-10
fix(mcp): share one pad-shape vocabulary between write_pcblib and update_pad

ande2407:
## Description

Two related pad-shape problems, both found while building a 1088-ball FCBGA footprint.

### 1. The two tools disagree on shape names (interop bug)

`write_pcblib` and `update_pad` each had their own `match` over shape strings, with different vocabularies:

| Value | `write_pcblib` | `update_pad` |
|---|---|---|
| `rounded_rectangle` | ✅ (**and the default**) | ❌ rejected |
| `circle` | ✅ | ❌ rejected |
| `Round` (any capitalisation) | ❌ rejected | ✅ |
| `rect`, `circular`, `oblong`, `octagon`, `rounded` | ❌ rejected | ✅ |

So the spelling `write_pcblib` documents **and defaults to** was rejected by `update_pad` — write a pad, then try to update it with the same shape string, and it errors.

Fixed by extracting one `parse_pad_shape` helper used by both, accepting the union case-insensitively and ignoring `_`/`-`, plus a shared `PAD_SHAPE_HELP` so both tools give identical guidance on error. **Input is only widened.**

### 2. The `shape` docs actively mislead for BGA/CSP

The description read:

> `round/circle (equivalent, for through-hole)`

BGA NSMD lands are round **and** SMD. Since omitting `shape` yields `rounded_rectangle`, a caller trusting that text — or omitting the field to shrink a large payload — silently converts every circular BGA land into a rounded rectangle. On a 1088-ball part that is 1088 wrong pads with no error raised.

The description now states the omit-default explicitly and names BGA as a round-land case. `update_pad`'s `shape` gains the same `enum` and wording; it previously advertised CamelCase values (`Rectangle, Round, …`) that didn't match the documented vocabulary.

### What is deliberately *not* changed

The default stays `rounded_rectangle`. Changing it would silently alter geometry for every existing caller that omits `shape`. It is now pinned by a test so any future change has to be deliberate rather than incidental.

## Type of Change

- [x] Documentation update

## Standards Checklist

- [x] Primitive types maintain backwards compatibility — accepted input is widened, never narrowed
- [x] Altium file format compatibility maintained (no writer change)
- [x] No breaking changes to MCP tool interfaces

## Testing

- [x] `pad_shape_vocabulary_is_shared_and_lenient` — every spelling from both former vocabularies resolves, plus casing/separator variants; `hexagon` still fails
- [x] `parse_pad_shape_defaults_to_rounded_rectangle` — pins the default and the explicit `round` opt-in
- [x] `update_pad_accepts_write_pcblib_shape_spellings` — regression for the interop gap; also asserts the error message lists the accepted spellings
- [x] `cargo test` — 718 lib + all integration suites (incl. `mcp_e2e` 36/36)
- [x] `cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean

## Notes

`docs/TOOLS.md` is unchanged — the generator renders top-level schema properties and both edited fields are nested (`pads[].shape`, `updates.shape`). The in-sync guard passes.

### DIFF
--- a/src/mcp/tool_definitions.rs
+++ b/src/mcp/tool_definitions.rs
@@ -221,7 +221,7 @@ impl McpServer {
                                                 "shape": {
                                                     "type": "string",
                                                     "enum": ["rectangle", "rounded_rectangle", "round", "circle", "oval", "octagonal"],
-                                                    "description": "Pad shape: rectangle (pin 1), rounded_rectangle (SMD default), round/circle (equivalent, for through-hole), oval, octagonal"
+                                                    "description": "Pad shape. Omitting this field yields rounded_rectangle, which suits chip/QFN lands but is WRONG for BGA/CSP: circular NSMD BGA lands must set \"round\" explicitly. rectangle is conventional for pin 1; round/circle are equivalent and used for both BGA lands and through-hole pads; also oval, octagonal. Matching is case-insensitive and ignores '_'/'-'."
                                                 },
                                                 "layer": { "type": "string", "description": "Layer name: Top Layer, Bottom Layer, Multi-Layer (default for SMD)" },
                                                 "hole_size": { "type": "number", "description": "Hole diameter for through-hole pads (mm)" },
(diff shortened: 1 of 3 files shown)


MatejGomboc:
Reviewed — unifying the two vocabularies into one lenient parser is exactly right, and pinning the rounded_rectangle default with a test means it can never change silently. The BGA guidance in the schema description is a bonus. Thank you @ande2407!

MatejGomboc (approved):

PR #296, opened 2026-08-11
fix(pcblib): write correct V7 layer id for Mechanical 17-32 (closes #282)

### LINKED ISSUE #282: bug: v7_layer_id() does not handle extended mechanical layers M17-M32
# Bug: v7_layer_id() does not handle extended mechanical layers M17-M32

## Description

The `v7_layer_id()` function in `src/altium/pcblib/writer.rs` does not handle extended mechanical layers M17-M32 (byte IDs 186-201, introduced in Altium Designer 18+). These layers fall through to the default `match` arm and return the `MultiLayer` fallback (`0x0102_0000`), which is incorrect.

This affects writing Track, Arc, Text, Fill, and Pad primitives placed on these layers. (Region and ComponentBody are unaffected because they use the string-based `v7_layer_token()`.)

## Reproduction Steps

1. Create a footprint with a Track or Pad on Mechanical 17 (byte layer ID = 186)
2. Read the footprint via MCP
3. Write it back via MCP
4. Inspect the written primitive — V7 layer ID will be `0x01020000` (MultiLayer) instead of the correct `0x01020011`

## Expected Behaviour

Primitives on M17-M32 should be written with correct V7 IDs:

| Layer | Byte ID | V7 ID |
|-------|---------|-------|
| Mechanical 17 | 186 | 0x01020011 |
| Mechanical 18 | 187 | 0x01020012 |
| Mechanical 20 | 189 | 0x01020014 |
| Mechanical 22 | 191 | 0x01020016 |
| Mechanical 28 | 197 | 0x0102001C |
| Mechanical 32 | 201 | 0x01020020 |

## Actual Behaviour

All layers M17-M32 are written as `MultiLayer` (`0x01020000`).

## Impact

- Footprints using M17-M32 lose layer information on round-trip
- Layers M17-M32 are commonly used custom mechanical layers

## Fix

Added handling for byte IDs 186-201 in `v7_layer_id()`:

```rust
if (186..=201).contains(&layer) {
    return 0x0102_0000 + (l - 169); // mechanical 17-32 (Altium 18+)
}
```

## Environment

- **OS:** Linux (WSL2)
- **Rust:** `rustc 1.93.1`
- **altium-designer-mcp version:** v0.1.0 (commit 3d7a019)

## Notes

- `Region` and `ComponentBody` are unaffected — they use `region_v7_layer_token()` which works via string tokens
- Layer M16 (byte ID 185) is already handled as Mechanical 16 (`0x01020010`)
- The fix is backwards-compatible — existing files are unaffected

## Related Files

- `src/altium/pcblib/writer.rs` — `v7_layer_id()` (line 589)
- `src/altium/pcblib/reader/mod.rs` — `layer_from_id()` (lines 357-488) — reading is already correct

ande2407:
Closes #282. Thanks @GorelkinAE — the diagnosis was correct and the file/line pointers were spot on.

## The fix

`v7_layer_id()` had no arm for byte IDs 186–201 (Mechanical 17–32, Altium Designer 18+), so they fell through to `_` and were written as the multi-layer fallback. Added the range, mirroring the existing mechanical 1–16 arm:

```rust
if (186..=201).contains(&layer) {
    return 0x0102_0000 + (l - 169); // mechanical 17-32 (Altium Designer 18+)
}
```

Verified the rest of the path already handles these layers and needed no change:

| Path | M17–M32 handled? |
|---|---|
| `layer_to_id` (M17–M32 → 186–201) | ✅ already |
| `Layer::parse("Mechanical 17")` | ✅ already |
| reader `layer_from_id` (186–201 → Layer) | ✅ already |
| `region_v7_layer_token` (Region / ComponentBody) | ✅ already |
| **`v7_layer_id`** (Pad, Track, Arc, Text, Fill) | ❌ **this PR** |

## Two corrections to the report

- **The wrong value was `0x0103_000F`, not `0x0102_0000`.** The `_` arm returns the multi-layer/fallback constant `0x0103_000F`. The conclusion (falls back to multi-layer) was right; only the constant was misquoted. The test asserts the real pre-fix value.
- **It wasn't only a round-trip defect.** `Layer::parse` accepts `"Mechanical 17"`, so a *fresh* write placing a pad or track on M17 was equally affected — not just read → write of an existing file. That widens the impact slightly.

## Why the test asserts on `v7_layer_id` directly

The MCP read tools resolve a primitive's layer from the **common-header byte**, which was always correct. So a read-back through `read_pcblib` looks clean while Altium — which renders from the V7 field — does not. That's the same masking behaviour as #124, and it's why a round-trip test would have passed on the broken code. The test therefore asserts the V7 field itself.

## Type of Change

## Standards Checklist

- [x] Primitive types maintain backwards compatibility — M1–M16 and all non-mechanical layers are untouched
- [x] Altium file format compatibility maintained (this restores it for M17–M32)
- [x] No changes to MCP tool interfaces

## Testing

- [x] `v7_layer_id_handles_extended_mechanical_layers` — checks the exact ids from the issue table; asserts mechanical N == `0x0102_0000 + N` contiguously and collision-free across **all 32** mechanical layers; asserts none equals the multi-layer fallback; cross-checks that the numeric id and `region_v7_layer_token` agree on the mechanical index (they are used by different primitive families, so they must not diverge)
- [x] Verified it **fails on the pre-fix code**: `left: 0x0103000F`, `right: 0x01020011`
- [x] `cargo test` — 716 lib + all integration suites (incl. `mcp_e2e` 36/36)

### DIFF
--- a/src/altium/pcblib/writer.rs
+++ b/src/altium/pcblib/writer.rs
@@ -589,6 +589,9 @@ fn v7_layer_id(layer: u8) -> u32 {
     if (57..=72).contains(&layer) {
         return 0x0102_0000 + (l - 56); // mechanical 1-16
     }
+    if (186..=201).contains(&layer) {
+        return 0x0102_0000 + (l - 169); // mechanical 17-32 (Altium Designer 18+)
+    }
     match layer {
         33 => 0x0103_0006, // top overlay
         34 => 0x0103_0007, // bottom overlay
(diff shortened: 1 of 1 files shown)


MatejGomboc:
Reviewed — the range arm mirrors the M1-M16 handling exactly and the test coverage is exemplary (the id/token agreement check is a lovely touch). Thank you @ande2407 for the fix and @GorelkinAE for the precise report!

MatejGomboc (approved):

PR #300, opened 2026-08-12
feat(pcblib): warn when pad copper overlaps and shorts nets

ande2407:
## Why

Overlapping pad copper merges into one net, so a footprint can pass every existing check while shorting pins together. Nothing caught it: `write_pcblib` and `validate_library` verify **integrity** (duplicate designators, zero-size primitives, non-finite coordinates), not **electrical sanity**.

This came from a real field failure. A TPA6130A2 WQFN-20 was built by reading TI's `(3.8)` land dimension as outer-edge-to-outer-edge when it is pad row **centre-to-centre**. Perimeter pads landed at 1.6 mm instead of 1.9 mm, so their inner edges (1.30) collided with a correctly sized 2.7 mm exposed pad (1.35):

```
pad inner edge  1.300 mm
EP half-width   1.350 mm   -> 0.050 mm overlap x 20 pads -> all 21 pads = one net
```

The library reported *"validation clean"* **twice** — before and after the bug was noticed — because the checks never looked at pad-to-pad geometry. On this part the EP-to-pad clearance is only 0.25 mm at best, so it is an easy error to make and an invisible one to catch.

## What

`Footprint::overlapping_pad_pairs()` in the pcblib layer, so both tool surfaces share one implementation and each reports through the channel it already has:

| Surface | How it reports | Note |
|---|---|---|
| `write_pcblib` | `pad_copper_overlap` entry in the existing `warnings` array, beside `silk_over_pad` | No response-schema change |
| `validate_library` | `severity: "warning"` issue | Lets you sweep **existing** libraries |
| `post_write_validation_pcblib` | *deliberately unchanged* | Would double-report inside `write_pcblib`'s own response |

Capped at 20 pairs plus a total-count summary, so a systematic error on a 1000-pad BGA can't bury the response.

## Deliberately excluded (warning, not error)

- **Pads sharing a designator** — same designator is the same net in Altium, so stacking them is a normal way to build a compound land
- **Pads with no layer in common** — a Top-only and a Bottom-only pad may sit at identical coordinates; `MultiLayer` shares both
- **Non-copper pads**

Rotated / non-rectangular pads are compared by axis-aligned bounding box, erring toward reporting: a false positive costs a glance, a missed short costs a board. Pads that merely **touch** are reported, since Altium merges those too (an EP of 2.6 mm on this part touches with zero gap and still shorts).

One design note: overlap is the **intersection rectangle's extent**, not penetration depth. My first implementation reported the latter and gave 0.47 mm in y for a pad lying wholly inside the EP's span — technically the separation figure, but misleading. The useful number is the pad's own 0.24 mm.

## Type of Change

## Standards Checklist

- [x] Primitive types unchanged — this is read-only analysis
- [x] Altium file format compatibility unaffected (no writer change)
- [x] No breaking changes to MCP tool interfaces — new entries in existing `warnings` / `issues` arrays

## Testing

- [x] `overlapping_pad_pairs_catches_undersized_qfn_land` — the field failure encoded: the real TPA6130A2 geometry at 1.6 mm centres yields exactly **20** pairs, each 0.05 × 0.24 mm; at the corrected 1.9 mm it yields **none**
- [x] `overlapping_pad_pairs_excludes_legitimate_constructions` — same-designator stacking, Top vs Bottom at identical coordinates, MultiLayer THT overlap (does warn), zero-gap contact (does warn), 0.2 mm clearance (does not)
- [x] `overlapping_pad_pairs_uses_rotated_bounding_box` — a pair that clears unrotated must collide at 90°
- [x] **End-to-end against the actual failing `.PcbLib`**: `validate_library` now returns `status: "warnings"` with 20 entries naming each pad pair and the 0.050 × 0.240 mm overlap; the corrected footprint returns `"valid"`. Both returned `"valid"` before this change.
- [x] `cargo test` — 723 lib + all integration suites (incl. `mcp_e2e` 36/36)

## Follow-up (not in this PR)

`update_primitive` covers track/arc/region/text/fill and pads have `update_pad`, but **vias have no update path** — moving one via currently means rewriting the whole footprint. Happy to send that separately.

### DIFF
--- a/src/altium/pcblib/mod.rs
+++ b/src/altium/pcblib/mod.rs
@@ -137,6 +137,77 @@ pub struct Footprint {
 }
 
 impl Footprint {
+    /// Finds pairs of pads whose copper overlaps on a shared layer.
+    ///
+    /// Overlapping copper merges into a single net, so a footprint can pass every
+    /// integrity check while shorting pins together — the classic cause being an
+    /// exposed thermal pad sized against a mis-read land dimension, which welds
+    /// the EP to every perimeter pad.
+    ///
+    /// Returns `(index_a, index_b, overlap_x, overlap_y)` in mm with `a < b`, where
+    /// the overlaps are the intersection rectangle's width and height.
+    ///
+    /// Deliberate constructions are excluded:
+    /// - pads sharing a designator (same designator == same net in Altium, so
+    ///   stacking them is a normal way to build a compound land),
+    /// - pads with no layer in common (a Top-only and a Bottom-only pad may sit
+    ///   at identical coordinates; `MultiLayer` shares both).
+    ///
+    /// Non-rectangular and rotated pads are compared by their axis-aligned
+    /// bounding box, so the test errs toward reporting: a false positive costs a
+    /// glance, a missed short costs a board. Pads that merely touch (zero gap)
+    /// are reported, since Altium merges those too.
+    #[must_use]
+    pub fn overlapping_pad_pairs(&self) -> Vec<(usize, usize, f64, f64)> {
+        /// Zero-gap tolerance. Altium stores coordinates in 2.54 nm units, so
+        /// anything at or below this is contact rather than clearance.
+        const TOUCH_TOL: f64 = 1e-6;
+
+        fn spans(pad: &Pad) -> (f64, f64) {
+            // Rotated pads use the AABB of the rotated rectangle.
+            if pad.rotation.abs() < f64::EPSILON {
+                return (pad.width, pad.height);
+            }
+            let radians = pad.rotation.to_radians();
+            let (cos, sin) = (radians.cos().abs(), radians.sin().abs());
+            (
+                pad.width.mul_add(cos, pad.height * sin),
+                pad.width.mul_add(sin, pad.height * cos),
+            )
+        }
+        fn shares_layer(a: &Pad, b: &Pad) -> bool {
+            let sides = |p: &Pad| match p.layer {
+                Layer::TopLayer => (true, false),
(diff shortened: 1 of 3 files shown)


MatejGomboc (approved):

MatejGomboc:
Reviewed in depth — this is excellent work, and the write-up (with the TPA6130A2 land-drawing misreading spelled out) is the clearest bug story this repo has had. The design calls are all the right ones: warning not error, same-designator and opposite-side exclusions, AABB erring toward reporting, and reporting the intersection extent rather than penetration depth. The three tests earn their keep, especially encoding the real 20-pair failure and its corrected counterpart.

Two things I found, both mine to fix rather than yours — I've prepared them as a follow-up branch (`fix/pad-overlap-cap`) that applies on top of this PR, so **nothing is blocking here**:

1. **The cap only exists on the write path.** `pad_copper_overlap_warnings` truncates at 20, but `validate_library` pushes one issue per pair with no limit. Since pairs are quadratic in pad count, sweeping a library that holds one systematically broken BGA can emit tens of thousands of issues into a single response — exactly the case your write-path cap was added for. Fix hoists the limit into the pcblib layer as `MAX_REPORTED_PAD_OVERLAPS` so both surfaces truncate alike, each with a summary line carrying the true total, plus a test (30 mutually overlapping pads = 435 pairs → 20 + summary).

2. **Cosmetic:** the `write_pcblib` message string has 22 literal spaces after its em dash, so it renders as `on Top Layer —                      overlapping copper merges into one net`.

Thank you @ande2407 — and thanks especially for chasing this from a real board failure rather than just filing it.

PR #301, opened 2026-08-12
feat(mcp): support vias in update_primitive

ande2407:
Follow-up flagged in #300.

## Why

Vias were the only footprint primitive with **no update path**:

| Primitive | Addressed by | Tool |
|---|---|---|
| Pad | `designator` | `update_pad` |
| Track, Arc, Region, Text, Fill | `index` | `update_primitive` |
| **Via** | — | **nothing** |

Moving a single via meant rewriting the entire footprint. I hit this fixing a TPA6130A2 land pattern: correcting one via position required re-sending 21 pads, 4 paste apertures, 13 tracks, a region, text and the 3D body — geometry that was already correct, re-transmitted for no reason.

## Why one arm, not an `update_via` tool

The pad/primitive split isn't arbitrary duplication — it keys on **addressing**. Pads are the only footprint primitive with a stable semantic identifier, so they get a designator-addressed tool. Vias carry no designator and are positional, exactly like tracks and arcs, so they belong in the index-addressed tool. A separate `update_via` would be a second door to the same room with nothing extra behind it.

**#295 is the cautionary case**: `write_pcblib` and `update_pad` each had an independently written pad-shape `match`, and they drifted until `rounded_rectangle` — the value one documented *and defaulted to* — was rejected by the other. Two surfaces mutating the same thing is how that happens. There's also an MCP-specific cost: every tool ships its schema on `tools/list`, so a redundant tool inflates what every agent reads on every connect.

## What

Supports `x`, `y`, `diameter`, `hole_size`, `from_layer`, `to_layer`.

Also repeats `parse_via`'s create-path invariants, because `update_primitive` bypasses `parse_via`: diameter and hole_size positive, hole smaller than diameter. Without that, an update could write a degenerate via the create path would have refused — the same reasoning behind the existing track-width and arc-radius re-checks.

The shared `create_rich_pcblib` fixture gains a via, so it again carries one of every primitive `update_primitive` can reach.

## Type of Change

## Standards Checklist

- [x] Primitive types unchanged
- [x] Altium file format compatibility maintained
- [x] No breaking changes — the `primitive_type` enum is widened, never narrowed

## Testing

- [x] `update_primitive_via_arm_persists` — updates all six fields, asserts each is reported in `changes` and persisted on reopen
- [x] `update_primitive_via_rejects_degenerate_geometry` — hole ≥ diameter, zero diameter, zero hole all rejected, **and the file is left untouched** in each case
- [x] `update_primitive_via_index_out_of_range` — bounds error names the primitive
- [x] `cargo test` — 723 lib + all integration suites (incl. `mcp_e2e` 36/36)
- [x] `docs/TOOLS.md` regenerated; the in-sync guard from #203 flagged the stale schema before I did

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -1096,5 +1096,5 @@ Update specific properties of a primitive (track, arc, region, or text) in a Pcb
 | `dry_run` | boolean | no | If true, show what would change without saving (default: false) |
 | `filepath` | string | yes | Path to the .PcbLib file |
 | `index` | integer | yes | Zero-based index of the primitive within its type array |
-| `primitive_type` | enum | yes | Type of primitive to update (one of: track, arc, region, text, fill) |
+| `primitive_type` | enum | yes | Type of primitive to update. Addressed by `index` into that primitive list. Pads are not here — they have a designator, so use update_pad. (one of: track, arc, region, text, fill, via) |
 | `updates` | object | yes | Properties to update (only specified properties are changed). Valid properties depend on primitive_type. |
--- a/src/mcp/tool_definitions.rs
+++ b/src/mcp/tool_definitions.rs
@@ -2044,8 +2044,8 @@ impl McpServer {
                         },
                         "primitive_type": {
                             "type": "string",
-                            "enum": ["track", "arc", "region", "text", "fill"],
-                            "description": "Type of primitive to update"
+                            "enum": ["track", "arc", "region", "text", "fill", "via"],
+                            "description": "Type of primitive to update. Addressed by `index` into that primitive list. Pads are not here — they have a designator, so use update_pad."
                         },
                         "index": {
                             "type": "integer",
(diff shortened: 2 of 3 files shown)


MatejGomboc (approved):

MatejGomboc:
Reviewed the diff — clean fill of a genuine gap. The via arm mirrors the existing track/arc/text/fill arms exactly (positional index, same out-of-range message shape), the dimension guard repeats `parse_via`'s create-path invariants so the two paths cannot drift, and the `from_layer`/`to_layer` loop is a tidy way to avoid writing the same block twice. Tests cover persistence, all three degenerate-geometry rejections with a check that the file is untouched, and the out-of-range case; adding the via to the shared `create_rich_pcblib` fixture was the right call. Nice catch keeping `docs/TOOLS.md` in step with the enum, too.

Thank you @ande2407 — that's three merged today.

Issue #302, opened 2026-08-12
Any interest in cutting a release with binaries?


ande2407:
Would you be open to cutting a tagged release with prebuilt binaries?

I've been building the exe and passing it to four colleagues here. We've been using it daily for Altium library work, everything from 0402 passives up to a 1088-ball BGA, and it's all compiled fine in Altium. It's felt solid for a while now.

Main reason I'm asking is that using it currently means having a Rust toolchain, which is a barrier for the PCB crowd this is aimed at. A downloadable binary would open it up quite a bit.

No worries if you'd rather keep it source only. This has been a lot of fun to contribute to either way, and I plan to keep at it.

MatejGomboc:
Thank you for asking — and genuinely, thank you for the contributions and for putting it through real work. Hearing that five of you have been using it daily on everything from 0402s up to a 1088-ball BGA, all compiling cleanly in Altium, is the most useful signal I've had about where this stands.

**Short answer: yes, and I'm planning it for the beginning of next week or thereabouts.**

I want one thing in place first. I'm not yet confident that every issue is resolved and every bug hunted down, so before there are versioned binaries in circulation I want test coverage up at **99%** — we're at 94.47% today. Today alone turned up a pad-overlap check with no cap on one of its two surfaces and a layer-id arm that silently mapped Mechanical 17-32 to the wrong layer, both in code that already looked finished. That's the pattern I want to stop seeing before people are running a downloaded exe: a file-format bug doesn't announce itself with a crash, it shows up as a corrupted library weeks later, and a binary tends to stick around and be trusted long after it's stale.

You're right about the barrier, though — needing a Rust toolchain filters out exactly the audience this is aimed at. The release plumbing already exists, so when it goes it will be a tag with prebuilt binaries rather than source-only. I'll leave this issue open to track it and will update here when it's cut.

If you'd like to help it land sooner, coverage contributions are the direct route. And please keep the fixes coming regardless — three of yours merged today (#300, #301, and the batch before), and the pad-overlap check in particular, chased down from a real board failure rather than filed and forgotten, is exactly the kind of thing that makes a release worth trusting.

ande2407:
Do you have a list of open items that are not covered today? if you do, please share. I can start working on your open items. For our uses right now, it covers everything we want it to do.

Happy to help contributing. We use it daily at this point. I just introduced it to another team, so doubling my user group.

MatejGomboc:
[...]

PR #304, opened 2026-08-14
test(library_ops): cover validate_symbol_json rejection paths

ande2407:
Toward the 99% coverage gate on #302.

`validate_symbol_json` gates every graphic primitive `write_schlib` accepts, and **none of its rejection branches had coverage** — a symbol whose pins or shapes were missing required coordinates reached the writer on the strength of guards nothing exercised.

This walks every family and every required field:

| Family | Fields checked |
|---|---|
| pins | `x`, `y`, `length` |
| rectangles, lines | `x1`, `y1`, `x2`, `y2` |
| arcs | `x`, `y`, `radius` |
| ellipses | `x`, `y`, `radius_x`, `radius_y` |
| labels | `x`, `y`, `text` |

Each omission must be rejected, and the message must name the symbol, the primitive index and the field — so a guard that gets dropped or mis-keyed fails here rather than silently passing bad input to the writer.

Also covered:
- the identifier fallbacks (`Unnamed`, `name='?'`, `designator='?'`) so a failure stays attributable when the caller omitted descriptive fields
- the accept path for **absent and empty** families, pinning the fact that a symbol need not carry every primitive type — guards against a future `is_empty()` check turning an ordinary pins-only symbol into an error

## Coverage

| | before | after |
|---|---|---|
| `library_ops.rs` (line) | 81.81% | **85.20%** |
| workspace (line) | 93.91% | **94.01%** |

Measured locally with `cargo llvm-cov --all-features --workspace`, matching the CI invocation.

## Testing

- [x] 3 new tests pass; `cargo test` green

Tests only — no behaviour change.

### DIFF
--- a/src/mcp/tools/library_ops.rs
+++ b/src/mcp/tools/library_ops.rs
@@ -1400,6 +1400,7 @@ mod tests {
 
     use crate::altium::pcblib::{Footprint, PcbLib};
     use crate::altium::schlib::{Pin, PinOrientation, SchLib, Symbol};
+    use crate::mcp::server::McpServer;
     use crate::mcp::tools::test_support::{
         create_test_pcblib, create_test_schlib, create_test_server, get_result_text,
         parse_result_json, test_temp_dir,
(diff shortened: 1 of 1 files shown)


MatejGomboc:
Reviewed line by line against `validate_symbol_json`. The case table matches the implementation exactly — all six families, every required field, in the same order the guards fire — and the two supporting tests (placeholder identifiers, absent vs empty families) cover the fallbacks and the negative case. Good call pinning `accepts_absent_and_empty_families`: that is the test that stops someone "tidying up" the guards into rejecting a pins-only symbol.

One weakness worth a one-line fix before merge. The assertion is:

```rust
assert!(err.contains("SYM") && err.contains(missing), ...)
```

For single-character field names this is close to tautological, because the message text itself contains those letters:

- `missing == "y"` — the word **Symbol** contains a `y`, so all four `y` cases (pins, arcs, ellipses, labels) pass no matter which field the implementation names.
- `missing == "x"` for **labels** — the message includes `(text='L')`, so the `x` in "text" satisfies it.

So five of the twenty-one cases would still pass if the guard reported the wrong field. Matching the quoted form pins it properly:

```rust
assert!(err.contains("SYM") && err.contains(&format!("'{missing}'")), ...)
```

Everything else looks right, and the multi-line `cases` formatting is rustfmt's doing rather than yours. Happy to merge once that assertion is tightened — or I can push the change if you would rather not round-trip. Thank you @ande2407.

MatejGomboc (dismissed):

PR #305, opened 2026-08-14
test(pcblib): raise reader/parsers.rs coverage from 76.9% to 93.8%

ande2407:
Toward the 99% coverage gate for the binary release (#302).

`src/altium/pcblib/reader/parsers.rs` was the worst-covered file in the repo. Round-tripping through the writer only ever produces the canonical, fully populated record, so none of the branches that matter for a *malformed or older* file were reached. These are the expensive ones: a bad record here does not crash, it yields a subtly wrong primitive that surfaces as a corrupted library weeks later.

**Tests only — the diff is 1001 insertions, 0 deletions.** No production line was touched.

## Coverage

`cargo llvm-cov --all-features --workspace --summary-only`

| | Lines | Regions | Functions |
|---|---|---|---|
| `parsers.rs` before | **76.89%** (247 missed / 1069) | 81.85% | 64.18% |
| `parsers.rs` after | **93.77%** (103 missed / 1653) | 94.85% | 80.33% |
| Workspace before | **93.91%** | 93.36% | 92.39% |
| Workspace after | **94.48%** | 93.85% | 93.15% |

The file's line denominator grows because `llvm-cov` counts the new in-file test module. The like-for-like number is the count of source lines with a zero execution count, all of which are in production code: **181 before, 47 after — 134 newly executed lines.**

## What is covered now

- **Minimum-length rejections** for every primitive: pad (<52 B), via (<31 B), track (<33 B), arc (<45 B), text (<25 B), fill (<37 B), region (<22 B), plus unreadable outer block-length prefixes for region and `ComponentBody`.
- **"Trailing field absent" template fallbacks** — a pad that stops at the mandatory 52 bytes, a via that stops at 31, a track/arc with no extended tail, a text with only 25 bytes. These assert the *Altium template constant*, not zero: a 0 mm thermal-relief conductor width or 0 mm plane clearance would make a pad flood-connect to an internal plane.
- **Unknown / out-of-range discriminants**: layer id, pad shape, hole shape, pad stack mode, via stack mode, text kind, and the full justification anchor table including the manual `0` and a corrupt `0xFF`.
- **Optional Block 5 layouts**: absent, present-but-too-short, legacy 289-byte (corner radius @288), canonical 596-byte (slot hole @262/@263/@267, corner radius @564), and the 576-byte full per-layer table including the shape-id-1 Round/RoundedRectangle disambiguation and the >100% radius clamp.
- **Region bounds guards**: parameter length past the end of the block, missing vertex count, a vertex count that overruns the block, and a hole-contour count that outruns the data.
- **`ComponentBody` outline truncation** (no parameter length, no vertex count, a half-present vertex) and the full `MECHANICAL1`–`MECHANICAL32` `V7_LAYER` range.

Each test carries a comment stating the concrete field consequence if that branch regressed.

### 1. Numeric text content is silently replaced by an unrelated `/WideStrings` entry

`resolve_text_content` treats **any all-digit content block** as a `/WideStrings` index:

```rust
if let Ok(index) = content.parse::<usize>() {
    if let Some(resolved) = ws.get(&index) {
        return resolved.clone();   // literal "1" becomes WideStrings[1]
    }
}
```

The writer emits `text.text` verbatim into block 1, and `read_pcblib` loads `/WideStrings` whenever the stream exists. So a footprint whose silkscreen text is literally `"1"` — pin-1 markers and numeric value legends are routine — reads back as whatever `WideStrings[1]` holds, and the next save persists the substitution. There is no signal to the user.

This is genuine silent data corruption, but fixing it is a behaviour change (it needs a real disambiguator — e.g. only resolving the index when the writer marked it as a reference). Pinned as a characterisation test, `numeric_text_content_is_swallowed_by_the_wide_strings_lookup`, so the fix shows up as a visible diff.

**No reachable panic was found.** Every direct index in this file is dominated by a length guard, every scalar read is bounds-checked, and the two allocation sites driven by a file-controlled count (`read_region_contour`, `parse_per_layer_data`) are both behind their length checks, so an attacker-chosen count cannot force an oversized allocation. `guid_string_from_bytes`'s `expect` is only ever reached through a checked `get(a..b)` of exactly 16 bytes.

### 2. The 99% gate is not reachable for this file without touching production code

Of the 47 remaining zero-count lines, **46 are provably dead** and cannot be covered by any input:

| Lines | Dead because |
|---|---|
| 114–115, 118–119 | pad x/y read failures, behind `geometry.len() < 52` |
| 141, 148 | `hole_size`/`shape` fallbacks gated on `len > 48` / `len > 49`, always true after the 52-byte guard |
| 406–407, 450–451 | per-layer size/offset `(0.0, 0.0)` fallbacks, behind the 320-/576-byte guards |
| 517–518 | via hole-size read failure, behind `len < 31` |
| 642–654 | track x1/y1/x2/y2 read failures, behind `len < 33` |
| 714–719 | arc x/y read failures, behind `len < 45` |
| 822–826 | text x/y read failures, behind `len < 25` |
| 1133–1143 | region vertex x/y read failures, behind `read_region_contour`'s `end` bounds check |
| 1191–1192 | region parameter-length read failure, behind `len < 22` |
| 1360–1370 | fill x1/y1/x2/y2 read failures, behind `len < 37` |

These are multi-line `ok_or_else` closure bodies whose preconditions the enclosing minimum-length check already guarantees. The 47th (line 394) is a lazy `tracing::trace!` argument, only evaluated with a TRACE-level subscriber installed.

That caps this file at roughly **95.6% of production lines** as written. Getting to 99% needs a production change — collapsing the redundant per-field error closures into the single length guard that already dominates them, or marking them unreachable. Happy to do that as a separate PR if you want it; it is deliberately out of scope for a tests-only change.

### DIFF
--- a/src/altium/pcblib/reader/parsers.rs
+++ b/src/altium/pcblib/reader/parsers.rs
@@ -1678,3 +1678,1004 @@ pub(super) fn parse_v7_layer(s: &str) -> Option<Layer> {
 ///
 /// The value is a tuple of (`stream_index`, `model_name`).
 pub type ModelIndex = HashMap<String, (usize, String)>;
+
+// =============================================================================
+// Tests
+// =============================================================================
+//
+// These drive the per-primitive parsers from hand-built byte buffers rather
+// than through the writer: the writer only ever emits the canonical, fully
+// populated record, so a round-trip cannot reach the short-buffer guards, the
+// "trailing field absent" fallbacks, or the unknown-discriminant arms. Those
+// are exactly the branches that decide whether a malformed or older file is
+// rejected loudly or loaded as a silently wrong primitive.
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Millimetre comparison tolerance — coordinates are quantised to 1 nm.
+    const EPS: f64 = 1e-9;
+
+    /// Wraps `payload` in the `[u32 len][payload]` framing every `PcbLib` record
+    /// block uses, so a test can hand-build a record without the writer.
+    fn block(payload: &[u8]) -> Vec<u8> {
+        let mut out = Vec::with_capacity(payload.len() + 4);
+        out.extend_from_slice(
+            &u32::try_from(payload.len())
+                .expect("fixture fits")
+                .to_le_bytes(),
+        );
+        out.extend_from_slice(payload);
+        out
+    }
+
+    /// Assembles a Pad record: the four leading blocks `parse_pad` reads before
+    /// the geometry block, the geometry block itself, and the optional Block 5
+    /// (the per-layer / size-shape block).
+    fn pad_record(geometry: &[u8], per_layer: Option<&[u8]>) -> Vec<u8> {
+        let mut data = Vec::new();
+        data.extend_from_slice(&block(&[0x01, b'1'])); // Block 0: designator "1"
+        data.extend_from_slice(&block(&[])); // Block 1
+        data.extend_from_slice(&block(&[0x04, b'|', b'&', b'|', b'0'])); // Block 2
+        data.extend_from_slice(&block(&[])); // Block 3
+        data.extend_from_slice(&block(geometry)); // Block 4: geometry
+        if let Some(per_layer) = per_layer {
+            data.extend_from_slice(&block(per_layer)); // Block 5
+        }
+        data
+    }
+
+    /// Writes the 13-byte `CommonPrimitiveData` header shared by every
+    /// primitive: layer @0, the unlocked flag word @1-2, and the `0xFFFF`
+    /// "no net / no polygon / no component" sentinels @3-8.
+    fn write_common_header(buf: &mut [u8], layer_id: u8) {
+        buf[0] = layer_id;
+        buf[1..3].copy_from_slice(&0x0004_u16.to_le_bytes()); // FlagUnlocked
+        buf[3..5].copy_from_slice(&0xFFFF_u16.to_le_bytes());
+        buf[5..7].copy_from_slice(&0xFFFF_u16.to_le_bytes());
+        buf[7..9].copy_from_slice(&0xFFFF_u16.to_le_bytes());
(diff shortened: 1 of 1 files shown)


MatejGomboc:
Reviewed. This is the best-argued PR the project has had, and the approach is exactly right: hand-built byte blocks reaching the truncation and fallback branches that a write→read round-trip can never touch, because the writer only ever emits the canonical fully-populated record.

Specific things I checked and liked:

- **The builders** (`block`, `pad_record`, `pad_geometry`, `via_block`, `region_block`) keep each test to the bytes that matter. That is what makes 25 tests readable rather than a wall of hex.
- **Asserting the Altium template constant, not zero**, on absent trailing fields. That distinction is the whole point — a 0 mm thermal-relief conductor width would flood-connect a pad to an internal plane, and a test asserting `0` would have locked the bug in.
- **The `0xFF` justification case and the >100% corner-radius clamp** — genuine hostile-input branches, not just line coverage.
- Your note on the line denominator growing because llvm-cov counts the in-file test module is correct, and the like-for-like figure (181 → 47 zero-execution source lines) is the honest way to state it. I measured the same effect independently.

**On the `/WideStrings` finding — you are right, and it is worse than a curiosity.** I traced it: the writer emits `text.text` verbatim into block 1 and the reader resolves any all-digit content through the table, so a footprint with a "1" pin-1 marker in a library carrying a `/WideStrings` stream reads back as an unrelated string, and read-modify-write persists it. Filed as #309 so it is tracked beyond this PR; the characterisation test is the right call and I have referenced it there. Fixing it needs a real disambiguator rather than a heuristic, so it should not hold up tests-only work.

The no-reachable-panic audit is appreciated too — I spot-checked `guid_string_from_bytes`'s `expect` and agree it is dominated by a checked 16-byte `get`.

No changes requested. Thank you @ande2407 — this is careful work.

MatejGomboc (approved):

PR #306, opened 2026-08-14
test(library_ops): cover delete/validate/export/import error paths

ande2407:
## Summary

Tests only — no behaviour change. Raises `src/mcp/tools/library_ops.rs` from the
lowest-covered large handler file toward the 99% gate in #302 by exercising the
rejection and failure branches that previously had none.

37 new tests in a `mod error_paths` block inside the existing colocated
`#[cfg(test)] mod tests`, using the standard `test_support` fixtures
(`create_test_pcblib`, `create_test_schlib`, `create_test_server`,
`get_result_text`, `parse_result_json`, `test_temp_dir`).

Every test carries a comment stating the concrete field consequence if the
branch regressed, and asserts on message content rather than just `is_error`,
so a guard that stops naming the offending component, index or field fails.

## Coverage

Measured with `cargo llvm-cov --all-features --workspace`, skipping the
`perf_tests` timing baselines (see note below).

| | Regions | Lines |
|---|---|---|
| `library_ops.rs` before | 83.41% | **81.81%** (266 / 1462 uncovered) |
| `library_ops.rs` after | 96.75% | **96.61%** (77 / 2269 uncovered) |
| Workspace before | 93.35% | **93.89%** |
| Workspace after | 94.11% | **94.67%** |

## What is now covered

- **`delete_from_pcblib` / `delete_from_schlib`** — path outside the sandbox,
  missing `component_names`, unreadable library on both paths, dry-run
  `not_found`, dry-run duplicate-name counting, non-dry-run `not_found`,
  backup failure (aborts without touching the library), and save failure
  (reports the failure while keeping the attempted deletions in `results`).
- **`validate_pcblib`** — sandbox rejection, unknown extension, unreadable
  library, degenerate tracks and arcs reported by index, region with fewer
  than three vertices.
- **`validate_schlib`** — empty library, symbol with no pins, pin with an
  empty designator.
- **`post_write_validation_pcblib` / `_schlib`** — reached through both delete
  paths; asserts pad, track, arc and region defects (PcbLib) and duplicate pin
  designators plus inverted rectangles (SchLib) are all reported.
- **`call_export_library` / `export_pcblib` / `export_schlib`** — sandbox
  rejection, unknown extension, unreadable library on both paths, and the
  compact-mode stack-mode downgrade (uniform full-stack pad is stripped and
  relabelled `simple`; a genuinely tapered stack keeps its per-layer sizes).
- **`call_import_library` / `import_pcblib` / `import_schlib`** — sandbox
  rejection, append against an unreadable existing library (asserts the file
  is left byte-identical), duplicate component name, out-of-range coordinates,
  unparseable footprint/symbol, failed save, missing `symbols` array, and the
  append success message carrying the running total.

`validate_symbol_json` is deliberately untouched — it is covered by #304.

## Bug found, not fixed

**`SchLib::open` accepts a `.PcbLib` file and reports it as an empty symbol
library.** `PcbLib::open` correctly detects the mismatch
(`Wrong file type: expected PcbLib, got SchLib (Schematic Library)`); the
SchLib side has no equivalent check and returns `Ok` with zero symbols.

Confirmed directly: saving a `PcbLib` under a `.SchLib` name and calling
`SchLib::open` on it succeeds with `len() == 0`.

For `delete_component` this is benign (nothing matches, so nothing is written),
but any append-style path that reads the existing library and then saves would
silently start from an empty library and overwrite the file. I have not fixed
it here to keep this PR tests-only; the wrong-type delete test was changed to
use a corrupt file instead so it still covers the intended branch. Happy to
open a separate issue.

## Note for the 99% gate

The 77 lines still uncovered in this file break down as:

- **40 lines** — `validate_symbol_json` (1220–1293), covered by #304.
- **20 lines** — the empty-component-name branches in `validate_pcblib`,
  `validate_schlib`, `post_write_validation_pcblib` and
  `post_write_validation_schlib`. These cannot fire: a component name is the
  OLE storage name, the writer refuses to save one that is empty, and the
  reader explicitly skips empty storage names (`read_io.rs`), so no file these
  functions can load ever yields an empty `name`.
- **16 lines** — the non-finite-coordinate branches for pads, tracks and arcs.
  Every coordinate these functions see comes from `to_mm(i32)`, which is always
  finite, so `!is_finite()` cannot be true for a library read from disk.
- **1 line** — the `unreachable!()` in `call_import_library`.

Excluding only the #304 block, the file sits at 98.37%. Reaching 99% would need
either those defensive checks removed, or the validators refactored to take an
in-memory `PcbLib`/`SchLib` so they can be unit-tested without a round-trip
through the file format. Flagging rather than deciding — happy to follow up
whichever way you prefer.

## Unrelated pre-existing issue

`perf_pcblib_save_100_footprints` fails under `cargo llvm-cov` instrumentation
(2.37 s against a 2 s threshold) on a cold-built worktree, which aborts the
coverage run before the report is produced. It passes normally under
`cargo test`. Both coverage numbers above were taken with
`-- --skip perf_` so they are directly comparable. Not touched here, but the
CI coverage job (`ci_main.yml` / `ci_pr.yml`) may hit the same threshold on a
slow runner.

## Gates

- `cargo fmt --all --check` — clean
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo test` — 766 lib + all integration tests pass (perf tests included)
- `library_ops` tests run 5× consecutively with no flakes

Refs #302

### DIFF
--- a/src/mcp/tools/library_ops.rs
+++ b/src/mcp/tools/library_ops.rs
@@ -2188,51 +2188,42 @@ mod tests {
             }
         }
 
-        /// Clears or sets the read-only attribute, ignoring failures on paths
-        /// that have already gone away.
-        #[allow(clippy::permissions_set_readonly_false)] // cleanup: the temp dir must stay deletable
-        fn set_readonly(path: &Path, readonly: bool) {
-            if let Ok(meta) = std::fs::metadata(path) {
-                let mut perms = meta.permissions();
-                perms.set_readonly(readonly);
-                let _ = std::fs::set_permissions(path, perms);
-            }
-        }
+        /// Makes the next save of `path` fail, on every platform, by occupying
+        /// the temp file `save_atomic` needs with a directory.
+        ///
+        /// Marking the library file read-only is the obvious approach and works
+        /// only on Windows: `save_atomic` writes `<stem>.<kind>.tmp` and renames
+        /// it over the target, and POSIX `rename` requires write permission on
+        /// the *directory*, not on the file being replaced — so on Linux and
+        /// macOS the save succeeded and the test did not. Making the directory
+        /// read-only instead would also block the pre-write backup, turning
+        /// these into duplicates of the backup-failure tests. Taking the temp
+        /// path fails `File::create` everywhere while leaving the backup free to
+        /// succeed, which is what these tests are actually about.
+        struct BlockedSave(PathBuf);
 
-        /// Makes a library file read-only for the guard's lifetime, so the next
-        /// write fails the way it does when Altium holds the file open or the
-        /// library lives on a read-only share. Write access is restored on
-        /// drop — including on the backup copies, which inherit the attribute
-        /// on Windows and would otherwise block temp-directory cleanup.
-        struct ReadOnlyFile(PathBuf);
-
-        impl ReadOnlyFile {
+        impl BlockedSave {
             fn new(path: &Path) -> Self {
-                set_readonly(path, true);
-                assert!(
-                    std::fs::metadata(path)
-                        .expect("metadata")
-                        .permissions()
-                        .readonly(),
-                    "test setup: could not make the library read-only"
-                );
-                Self(path.to_path_buf())
+                // Mirrors the extension `save_atomic` is called with for each
+                // library kind, so the blocked path is the one it will use.
+                let tmp_ext = if path
+                    .extension()
+                    .is_some_and(|e| e.eq_ignore_ascii_case("schlib"))
+                {
+                    "schlib.tmp"
+                } else {
+                    "pcblib.tmp"
+                };
+                let temp = path.with_extension(tmp_ext);
+                std::fs::create_dir_all(&temp)
+                    .expect("test setup: could not occupy the save temp path");
+                Self(temp)
             }
         }
 
-        impl Drop for ReadOnlyFile {
+        impl Drop for BlockedSave {
             fn drop(&mut self) {
-                let prefix = self.0.to_string_lossy().into_owned();
-                let parent = self
-                    .0
-                    .parent()
-                    .unwrap_or_else(|| Path::new("."))
-                    .to_path_buf();
-                for entry in std::fs::read_dir(parent).into_iter().flatten().flatten() {
-                    if entry.path().to_string_lossy().starts_with(&prefix) {
-                        set_readonly(&entry.path(), false);
-                    }
-                }
+                let _ = std::fs::remove_dir_all(&self.0);
             }
         }
 
@@ -2490,7 +2481,7 @@ mod tests {
             let server = create_test_server(dir.path());
             let path = dir.path().join("Locked.PcbLib");
             create_test_pcblib(&path);
-            let _readonly = ReadOnlyFile::new(&path);
+            let _blocked = BlockedSave::new(&path);
 
             let result = server.call_delete_component(&json!({
                 "filepath": path.to_string_lossy(),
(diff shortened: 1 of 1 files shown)


MatejGomboc:
Reviewed. Broad, well-targeted coverage of exactly the branches I had flagged as the bulk of the gap — sandbox rejection, unreadable libraries, dry-run vs real, backup failure, failed write, duplicate names, out-of-range coordinates. The `set_readonly` / `block_backup_paths` helpers are a neat way to reach the backup-and-save failure paths without mocking the filesystem, and asserting the file is left **byte-identical** after a rejected append is the right assertion rather than merely checking the error.

Two things I particularly valued:

- Scoping out `validate_symbol_json` because #304 owns it — that avoids the duplicate-test collision I asked contributors to watch for.
- The closing breakdown of what remains uncovered, including the 20 empty-component-name lines you judge unreachable. That is the kind of note that stops the next person re-deriving it.

**On the `SchLib::open` finding — confirmed independently before taking it at face value.** I saved a real `PcbLib` (one footprint, one pad) under a `.SchLib` name and opened it: `Ok`, `symbol_count == 0`.

Worth adding to your write-up: there *is* already a wrong-file-type test on the SchLib side and it passes, which is why this went unnoticed. It hand-builds a hybrid — a SchLib-format length-prefixed `FileHeader` whose text contains the PcbLib header string. That shape is detected; a real PcbLib's binary 53-byte `FileHeader` never parses as that string, so the guard never fires on the file shape that actually occurs. The existing test was asserting the guard works against a file no tool produces.

Filed as #310, with your reasoning about append-style paths and a note to check whether `SchLib::open` returns "silently empty" for other unreadable inputs too. Your decision to switch the delete test to a corrupt file so it still covers the intended branch was the right one for a tests-only PR.

No changes requested. Thank you @ande2407.

MatejGomboc (dismissed):

MatejGomboc:
Heads-up on why CI went red here — **the tests are correct; they found a real bug in production code.**

Your `reports_a_library_it_cannot_read` tests assert that a failed read names the library by file name only, which is the project's stated path-sanitisation rule. They fail because `redact_absolute_paths` does not actually hold that rule. Chasing it down turned up three separate leaks, all now fixed in #312:

1. **Paths containing spaces** — the path body excluded whitespace, so a match stopped at the first space and the rest of the directory tree stayed in the message. This is how I first hit it locally (`…\embedded society\…`).
2. **Quoted paths on Unix** — the Unix pattern required a whitespace boundary, but every tool response is JSON, so a path is reached as `"filepath": "/home/…"`, preceded by a quote. Those were not partially redacted, they were **not redacted at all** — which is exactly what your ubuntu run reported, printing the whole runner workspace path.
3. **JSON-escaped separators and verbatim prefixes** — redaction runs on the already-serialised JSON body, where Windows separators arrive doubled, and `canonicalize` returns verbatim `\\?\C:\…` paths. Both broke matching, producing `<path>?\C:Corrupt.PcbLib`.

So the value of this PR is higher than the coverage number suggests: it caught a live path-disclosure bug on every platform, in the project's own defence-in-depth choke point.

Verified locally that with #312 applied, **all 61 error-path tests here pass** and the full suite is green at 812 lib tests. Nothing needs changing on your side — once #312 lands, merge `main` into this branch (or a maintainer will) and CI should go green.

No changes requested. Thank you @ande2407 — genuinely good work.

MatejGomboc:
Conflict resolved and pushed to this branch.

`main` had gained #304's `validate_symbol_json` tests, which were appended to the same `mod tests` block as this PR's `mod error_paths`, so git could not tell the two additions apart. Both are kept — nothing was dropped from either side.

Also merged in #312, which fixes the three path-disclosure leaks your tests exposed. Verified locally on the merged tree:

- `cargo test` — **815 lib tests pass**, plus every integration suite, including all 61 tests in `error_paths` and #304's three.
- `cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` both clean.

CI should now go green.

PR #348, opened 2026-08-15
feat(pcblib): model the barcode sizing block

MatejGomboc:
**Corrects #347**, which recorded this as blocked. Both of its claims were wrong.

## What I got wrong

I said AD24 exposes no setter for the barcode keys. **Our own `GenerateSamples.pas` has
been authoring them** since the barcode fixture was added:

```pascal
Txt.BarCodeKind       := eBarCode128;
Txt.BarCodeFullWidth  := MilsToCoord(400);
Txt.BarCodeFullHeight := MilsToCoord(100);
```

I also said our two text records carried identical template bytes at the candidate
offsets. They differ at **31** offsets, and `@137` already held that authored 400 mil.

The root cause was a bad heuristic I had been leaning on and had even published: that a
missing `Set*` counterpart in the RTTI proves a property is unauthorable.
`SolderMaskExpansionFromHoleEdge` and `BarCodeKind` both lack one and both set fine.

**What actually holds:** whether the name appears in **`Advpcb.dll`**, the native Delphi
engine. Names present only in the `Altium.*.dll` **.NET** assemblies do not resolve in
DelphiScript — `TextJustification` is one, and assigning it aborts the whole script
compile with *Undeclared identifier*. That is how I found this.

## Six keys, each offset derived

Authored a second barcode with every field different and matched the authored value to the
byte carrying it:

| offset | field | BC128 → BC2 |
|---|---|---|
| @137 | FullWidth | 400 → 600 mil |
| @141 | FullHeight | 100 → 150 mil |
| @145 | XMargin | 20 → 30 mil |
| @149 | YMargin | 20 → 40 mil |
| @157 | Kind | `1` = Code128 (`0` on a non-barcode) |
| @161-224 | FontName | `'Arial'` → `'Courier New'`, UTF-16LE, null-padded |

The font field is the odd one: UTF-16LE in a record where every other string is
Windows-1252.

Read, written, exposed on the write schema, and golden-asserted against Altium-authored
bytes. The block is surfaced **only** for a barcode, so a plain text still replays the
template untouched — asserted against the `INV` record, which is what stops the block
being read unconditionally out of bytes every text record happens to carry.

## Still open

`MinWidth`, `Inverted`, `ShowText`, `RenderMode`. `@153` reads 39604/88235 against an
authored 5 mil, so Altium computes it rather than storing ours; the booleans produced no
isolated flip. They need a run varying one field at a time — no longer "blocked", just
not done.

## Testing

1040 tests green, `cargo fmt --all --check`, `clippy -D warnings`, markdownlint clean,
pyaltiumlib oracle 13/13. Docs corrected in the audit (including the corrected heuristic,
written down so it does not get re-derived wrongly) and the fixture map.


MatejGomboc:
Pushed a second commit: **`DrillType` is resolved as a negative too.**

#346 recorded it as blocked, needing a press-fit pad from a real library. That rested on the same faulty `Set*` reasoning this PR corrects, so I retested it: `Pad.DrillType := 1` compiles and runs, and the saved pad is then byte-identical to a plain through-hole pad apart from its coordinates. AD24 keeps that classification somewhere other than the library record.

So no external file is needed and there is nothing to model — it joins `TearDrop`/`UserRouted` as a documented negative rather than sitting on the worklist. Both requests I made of @ande2407 in #302 have now dissolved; I have said so there.

MatejGomboc:
Third commit: **barcode `Inverted` (@159) and `ShowText` (@225)** are now modelled too, isolated with BC3/BC4 differing from BC2 in exactly one field each.

That makes **eight of the ten** barcode keys modelled. The last two are negatives rather than gaps:

- `MinWidth`@153 reads 39604/88235 against an authored 5 mil — Altium computes it from the content and width rather than storing the request.
- `RenderMode`: a BC5 varying only that field moved no byte except @115, which reads 4/3/2/1 across the four barcodes in creation order — an ordinal, not the property. BC5 was removed rather than left asserting nothing.

1040 tests green, clippy and markdownlint clean.

PR #378, opened 2026-08-16
docs: the five damaged fixtures trace to AD's reader; resave is negative

MatejGomboc:
Answers "can the manual fixture fix be automated?" with a definitive **no** — and finds the root cause while trying.

## The idea (run 4)

The damaged records carry the *true* names in their `%UTF8%` twins, widened through this machine's ANSI code page. So: have AD **open the golden and resave it** — its reader recovers the names from the twins, its writer re-emits consistent records, and no script string ever holds a name (bypassing the engine defect entirely). One new `ResaveRun` procedure in the existing `GenerateSamples` project, same launcher/bridge/watcher, `ProcName="GenerateSamples>ResaveRun"`.

## The result

A **fourth mangling variant, worse than the input**: replacement characters appear (Bengali degrades to mostly `U+FFFD`; `𠮷野` comes back real but trailed by garbage; Javanese still shifted). 

## What that proves

**AD's reader is the broken component.** It cannot losslessly decode these five byte sequences even from its own file. One defect now explains all four runs: the script engine feeds literals through the same decode (run 1's shifted records), and every open+save cycle degrades the five further (run 4). The writer side is faithful throughout — the 48 working symbols in the same file prove it.

And it confirms the remaining cure works: typing the five names once in the **AD UI** bypasses the decode entirely (UI input → real wide string in memory → faithful writer). The repo never re-opens goldens in AD, so reader-side lossiness never touches the committed file again.

## Changes

- `GenerateSamples.pas`: `ResaveRun` procedure, kept with the negative recorded in its doc comment (it is the tool a future decode investigation reaches for first); the i18n `DOCUMENTED NEGATIVE` unchanged.
- `COVERAGE_AUDIT.md`: the five-fixture entry now records all four runs and the reader-side root cause.
- Goldens untouched; full suite green (12 suites); preflight clean.

### DIFF
--- a/docs/COVERAGE_AUDIT.md
+++ b/docs/COVERAGE_AUDIT.md
@@ -16,13 +16,17 @@ still loses is listed in that test's `KNOWN_DEFECTS` and described here; the two
 the same list, so an entry leaves both together.
 
 **Five i18n fixture symbols are internally inconsistent** (`_JV`, `_BN`, `_CR`, `_IU`,
-`_SB`) — and scripted regeneration is a **verified negative** (three runs, 2026-08-16).
-Source literals: the engine mis-decodes exactly these five sequences (storage name correct,
-records shifted). Wide `Chr($A997)`: truncates to the low byte — the engine's strings are
-ANSI. UTF-8 byte `Chr($EA)+…`: storage names and `SectionKeys` come out byte-perfect but
-every text record double-widens. No scripted construction remains; the cure is renaming
-the five once by hand in the AD UI. Until then `tests/golden_fidelity.rs` excuses exactly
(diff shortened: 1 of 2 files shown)


MatejGomboc:
Pushed a second commit: the full 84-symbol diff of the resave experiment found two more casualties — `ꆈꌠ_YI` (pin/label/parameter texts degrade to mojibake after one load+save) and `PARAMS` (`100nF` parameter value comes back as `*`). 77/79 non-damaged symbols survive a cycle, but that makes open-and-save a lottery, not a maintenance tool. The samples README now carries the warning, and the manual fix for the five damaged fixtures will be authored as a fresh `manual/` library instead of editing the golden.

PR #379, opened 2026-08-16
test: take the tool layer to 98-99%, and two findings about the 99% gate

ande2407:
Picks up the coverage work from #302. Eleven files, 4,248 lines of tests, no production code touched.

**94.47% → 97.88%** overall (1,792 uncovered lines → 843), measured with the same command CI runs.

| File | Before | After |
|---|---|---|
| `mcp/tools/parsing.rs` | 90.86% | 99.69% |
| `mcp/tools/read_write.rs` | 92.64% | 99.62% |
| `mcp/tools/components.rs` | 92.37% | 99.51% |
| `mcp/tools/batch.rs` | 95.18% | 99.38% |
| `mcp/tools/compare.rs` | 90.65% | 99.38% |
| `mcp/tools/maintenance.rs` | 92.75% | 98.90% |
| `mcp/tools/query_update.rs` | 95.56% | 98.87% |
| `altium/pcblib/read_io.rs` | 80.70% | 90.00% |

Plus `library_ops.rs`, `validation.rs`, `server.rs` and `pcblib/reader/mod.rs`.

## What the tests cover

Mostly item 2 from your list — the tool-layer error paths. Each tool guards its arguments, dispatches on the file extension into a PcbLib and a SchLib implementation, and each implementation guards the read, the lookup and the write. The existing fixtures exercised the happy path and usually only the PcbLib half, so the guards were largely untested. There is now a case per missing argument, per path outside the allow-list on every path argument a tool takes, both dispatch failures, an unreadable library on each side that gets read, each lookup failure, and the write failure.

The write failure needed a mechanism that does not depend on running unprivileged: `validate_path` refuses a missing parent directory before save is ever reached, so the tests mark the library read-only instead. It still opens and still backs up — both only read it — so the save is what fails, which is the branch every mutating tool funnels through.

Some of item 1 and item 3 too:

- **`layer_from_id`** is the reader-side twin of the writer table #353 covered, and the old test sampled about 30 of its 100 mapped ids. It now walks every one and demands 100 distinct layers, which is the check that would have caught #282 rather than a spot-check that happened to miss it.
- **Stream parsing under corruption** — the UniqueID length-prefixed record scan (length past the buffer, zero length, past the 10 kB cap, trailing partial prefix, unusable body) and WideStrings dropping malformed entries individually.
- **`read_io.rs`** was the worst-covered file in the crate. The legacy `FileHeader` fallback is only reachable from an older library, so rather than wait on a sample the tests build the compound document directly — a `/FileHeader` stream and nothing else drives every header path, including a SchLib opened as a PcbLib being refused by name.

A few behaviours are pinned because a wrong reading would be silent rather than loud: a pad's layer defaulting from whether it is drilled, per-layer stack entries falling back per entry so one bad shape does not lose the other 31, an already-matched primitive not being claimed twice by the diff, and a dry-run merge simulating renames without creating the target.

## Two findings about the 99% gate

Both turned up while chasing the number, and I think they change what the gate can mean.

**1. A large share of what remains is unreachable by construction.** The readers guard a block's length and then read inside it, which makes the error branches beneath the guard dead code. `parse_track` checks `block.len() < 33` and then reads i32s at offsets 13, 17, 21, 25 and 29 — the last reaching byte 33 exactly — so all five `ok_or_else` arms are unreachable. `read_region_contour` checks `props_block.len() < end` where `end` is precisely the extent its vertex loop then reads. `parse_pad` does the same, plus `if geometry.len() > 48` / `> 49` fallbacks that cannot fire because an earlier check already required ≥ 52. That is roughly 44 of `parsers.rs`'s remaining 111 lines, and the pattern repeats through the via, arc, text and fill parsers.

Same story elsewhere: `main.rs` is 58 lines of entry point and async run loop, `server.rs` has ~40 uncovered lines inside its own test module, and `library_ops.rs`'s `!x.is_finite()` checks cannot fire on a round-tripped file because NaN saturates to 0.0 on write.

**2. The denominator moves.** CI measures with test code included, so new tests raise the total they are trying to lift. Concretely: `server.rs` went from 79 uncovered to 82 after I added five passing tests — the production lines got covered, and the new module's own assertion-failure branches more than offset them.

Together those put the reachable ceiling somewhere around **98.3%**, not 99%. Worth deciding whether to exclude test code from the metric (`--ignore-filename-regex`), set the gate against reachable code, or drop the guard-then-read defensive arms that can never run. Happy to do any of those — I did not want to touch production code in a test PR.

## Checks

`cargo build`, `cargo test` (1,013 lib tests + every integration suite, 0 failures), `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`. No markdown changed. No CHANGELOG entry — nothing here is user-facing.


MatejGomboc (dismissed):
Reviewed with a full local verification, since your branch predates a busy day on `main` (#370–#378 merged since your base — the whole round-trip fidelity chapter):

- clean merge onto tonight's `main`, `cargo build`, **all 12 test suites green**, `clippy --all-targets -D warnings` clean;
- additions verified strictly inside `#[cfg(test)]` modules — the "no production code" claim holds structurally, not just by intent;
- every `set_readonly(true)` is paired with a restore, so Windows tempdir cleanup can't trip;
- the read-only write-failure mechanism is the right trick — it fails exactly the branch every mutating tool funnels through, unprivileged;
- the exhaustive `layer_from_id` walk demanding 100 distinct layers is precisely the shape of test that catches the next #282.

Worth noting out loud: your error-path tests survived eight subsequent PRs *unchanged*, including ones that altered writer behaviour (`MODEL.*` on extruded bodies, `AllPinCount`, SectionKeys). That's evidence they pin contracts rather than incidentals.

On the two findings — both are real and change what the gate can mean:

1. **Unreachable guard-then-read arms**: agreed on the diagnosis, but I'd argue against deleting them. The `ok_or_else` arms under a length guard are insurance against the guard and the reads drifting apart in a future refactor — exactly the class of silent corruption this repo has been hunting all week. Dead under today's guards, load-bearing the day someone edits an offset. Measuring around them beats removing them.
2. **The moving denominator**: agreed, and excluding test code from the metric (`--ignore-filename-regex` on the llvm-cov invocation) is the standard cure — it also makes finding 1 mostly moot numerically.

The gate itself (99%-of-what) is Matej's call as release owner — I've put the options in front of him, and #302 will be updated in place once decided. The tests themselves are merge-ready as-is. Thank you — this is exactly the second pair of hands the coverage climb needed.

MatejGomboc:
CI failure diagnosed and fixed — pushed `64fb2a6` to your branch (thanks for leaving maintainer edits on).

**What happened:** the read-only-file trick fails the save only on Windows. The save path is atomic — write a sibling temp file, then rename it over the target — and on Unix rename permission is governed by the **directory**, not the file, so the read-only target was happily replaced and the expected error never came: every failed-write test passed on Windows and panicked on Linux/macOS.

**The fix:** `set_readonly` keeps its name and all call sites, and splits by platform — Unix locks the parent directory (`0o555`), which refuses the temp-file creation the atomic save needs; Windows keeps the read-only file attribute, which is what blocks the replace there. Both funnel into the same save-failure branch these tests exist to cover, and your plain `is_error` assertions hold at either failure point.

Verified on Windows locally (1,013 lib tests green, clippy `-D warnings` clean); CI will now speak for Linux and macOS. Nice mechanism, by the way — it just needed to know which platform enforces read-only where.

MatejGomboc:
Round two pushed (`bee0d8f5`): my directory lock was still wrong on Unix — it failed the **backup** first (the `.bak` lands in the same directory), while three of your tests rightly assert on the save branch specifically. The fix drops permissions entirely and squats a directory on the deterministic temp path `save_atomic` must create (`<name>.pcblib.tmp`) — the same `BlockedSave` mechanism `library_ops.rs` already uses for exactly this reason. `File::create` over a directory fails identically on every platform and the backup proceeds untouched, so there is no cfg split: a green run on one platform now exercises the same code all three run. Helper renamed `set_readonly` → `block_save` to match what it actually does. Windows local: 1,013 lib tests + all 12 suites green, clippy clean.

PR #383, opened 2026-08-16
refactor(pcblib): report truncation per field instead of by block length

ande2407:
TODO §B item 2 — the guard-then-read rework, picking up the offer from #302.

Each `PcbLib` parser opened with a constant length guard and then read its fields through bounds-checked accessors that could no longer fail, because the guard had already proved the block long enough. Every one of those `ok_or_else` arms was dead code: the checks that actually protect the parse could never run, so they could never be tested either.

Dropping the guards makes each read the thing that reports its own failure:

```
before: Track block too short: 32 bytes, expected at least 33
after:  Parse error at offset 29: failed to read Track width
```

Rejection behaviour is otherwise unchanged — the same truncated blocks are still refused, and the message now names the field rather than a byte count the caller has to map back to one.

Fields previously read with a bare index (`block[0]`, the via's layer span at `block[29]`/`[30]`) are now read fallibly, so no truncated block can panic where the guard used to return an error. Data-driven length checks — a vertex count or parameter-string length read out of the block itself — are a different thing and stay.

## Two things worth your eye

**A boundary moves, in Pad.** Its guard demanded 52 bytes, but nothing lives at 50 or 51: the last field the parser names is the shape byte at @49. The `if geometry.len() > 48` / `> 49` fallbacks beneath the guard were vestigial, so hole size @45 and shape @49 are now required reads and the accept boundary is 50. A 50- or 51-byte geometry block parses where it used to be refused — it carries a complete pad, and refusing it was an artefact of a hand-picked constant rather than a judgement about the data. Both sides of the boundary are pinned by tests. If you'd rather keep the old 52 exactly, it needs a magic constant back, since 52 isn't a field boundary — say the word.

**One dead branch survives, deliberately.** Arc's two angle doubles at @25/@33 carry `unwrap_or` defaults (0 and 360, a full circle) rather than failing, and the required width at @41 sits past both — so anything long enough to reach the width has already supplied them. Their defaults stay unreachable. That's noted in the arc test rather than worked around.

## Tests

The seven "block too short" tests become per-field tables, keeping each one's note on what a silently-accepted truncation would cost — a pad that solders to nothing, a via that stitches nothing, a track absent from the Gerber output. Each case asserts the field named in the message, so the table doubles as the parser's required-field list.

`parsers.rs`: **93.59% → 98.14%** (111 uncovered lines → 33). Project total on the production metric from #381: **98.02% → 98.21%**.

`cargo build`, `cargo test` (1,014 lib tests + every integration suite, 0 failures), `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`. The golden-fidelity round-trips passing is the load-bearing check here: real Altium files still parse identically.

### DIFF
--- a/src/altium/pcblib/reader/parsers.rs
+++ b/src/altium/pcblib/reader/parsers.rs
@@ -1,4 +1,24 @@
 //! `PcbLib` reader: per-primitive binary parsers (pad/via/track/arc/text/region/fill/component-body).
+//!
+//! # Truncation is caught per field, not by an upfront length guard
+//!
+//! Each parser reads its fields through bounds-checked accessors and reports the
+//! first one that runs off the end of the block, naming that field. There is no
+//! "block too short" pre-check, and deliberately so: a guard that asserted the
+//! block was long enough for every field made each read's own error arm
+//! unreachable, so the checks that actually protect the parse could never be
+//! exercised — and a reader whose error handling cannot be tested is a reader
+//! whose error handling is assumed rather than known.
+//!
+//! Rejection behaviour is unchanged: the same truncated blocks are still
+//! refused, and the message now names the field that could not be read instead
+//! of quoting a byte count the caller has to map back to a field themselves.
+//!
+//! The consequence for anyone editing a parser: **read every field through
+//! `read_*`, `.first()` or `.get()`**. A bare `block[n]` panics on a truncated
+//! block where the old guard would have returned an error. Length checks that
+//! are *data-driven* rather than constant — a vertex count or a parameter-string
+//! length read out of the block itself — are a different thing and stay.
 
 #[allow(clippy::wildcard_imports)] // tightly-coupled reader split
 use super::*;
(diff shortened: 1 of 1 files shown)


MatejGomboc (approved):

MatejGomboc (approved):
Reviewed with the full local protocol:

- merged onto tonight's `main`: **all 12 suites green**, clippy `-D warnings` clean;
- merged onto `main` **plus the open #384** (which reworks `parse_component_body` in the same file): git auto-merges cleanly and — the part that matters — the combined tree passes the full suite and clippy too. So the two can land in either order with no rebase pain;
- the per-field failure tables are exactly the right replacement shape — the error now names the field, and the table doubles as the parser's required-field spec.

**Verdict on the Pad boundary: your 50 stands.** The old 52 was a hand-picked constant, not a field boundary — nothing we name lives at @50/@51, every golden pad is comfortably longer, and accepting a complete-but-short block matches the reader-tolerance stance used everywhere else in this codebase ("a partial read beats refusing the footprint"). The one scenario where 52 would have been safer — a future field at @50 silently defaulting instead of rejecting — is better handled the day such a field is *found*, with a fixture, than pre-emptively with a magic number. Your tests pinning both sides of the boundary are what make that a safe position.

The deliberately-surviving Arc dead branch (angle defaults unreachable past the width read) documented in the test rather than contorted around — good judgement, that's the honest form.

And the headline: `parsers.rs` 93.59% → 98.14%, project **98.21%** on the production metric. The 99% v1.0.0 gate is now within a single PR's reach. Thank you — this closes the item with your name on it from #302, and does it better than the sketch suggested.

PR #385, opened 2026-08-16
test: close most of the remaining coverage gap, and where the floor is

ande2407:
TODO §B item 3, following #379 and #383. Seven passes, 16 files, tests only.

**98.44% → 98.74%** on the production metric from #381 (621 uncovered lines → ~515). With #379 and #383 that's 1,792 → ~515 across the three PRs.

## What this covers

**Encoders and their limits.** Every cap in the SchLib writer guards against silent corruption rather than a loud failure: the record length is a 24-bit field and the pin strings are Pascal short strings, so an oversize value truncates and every record after it is read from the wrong offset. Covers the 16 MiB record cap, the i16 range on pin coordinates and owner part id, and the 255-byte cap on all six pin strings — including that the cap is measured in *encoded* bytes, so 255 non-ASCII characters fit where a UTF-8 length check would wrongly refuse them.

**The `V7_LAYER` vocabulary.** Altium resolves a Region's or ComponentBody's layer from this token, not the header byte, and an unresolvable token leaves the primitive on Top Layer — so a wrong entry silently moves copper to the other side of the board, which is what #124 and #282 were. Walks the vocabulary and checks no two layers collide.

**Stop-on-corruption contracts.** The record dispatch (all eight types), the model index, and the `/Storage` walk are each length-prefixed and read back to back, so a parser that fails has lost its place; continuing would invent primitives that were never in the file. One case per arm asserts each stops instead. The storage walk gets the distinction that matters too: a payload that fails to inflate costs its own image but leaves the framing intact, so later entries still arrive.

**Library read paths, via built compound documents.** The golden fixtures are valid by construction, so the header and component-walk failures needed malformed libraries — these build the CFB directly. Reading a footprint library as a symbol library must fail rather than look empty (an append-style caller seeing "zero symbols" would save an empty library over the file); a component the header names but the file lacks costs that symbol and no more; and a footprint's canonical name comes from PATTERN, not the 31-character storage name, with or without the block's length prefix.

**Per-field range checks.** The symbol families are validated one field at a time, so covering one rectangle corner said nothing about the other three. Every remaining coordinate now has a case.

**Serde defaults**, which fire on the read-modify-write path where the writer omits anything at its default. The via ones matter most: a zero thermal-relief gap flood-connects the via into a plane and makes the board unsolderable. And SchLib anchors text bottom-left rather than the shared enum's MiddleCenter, so taking the shared default would shift every label on save.

**Pin aux streams**, each keyed by pin ordinal — a malformed entry has to be skipped alone, or one pin's offset or name lands on a different pin and the symbol opens subtly wrong with nothing to explain it.

## Where the floor is

Worth putting in front of you before more time goes into this. Of the ~515 remaining, roughly **125 cannot be reached by any test**, confirmed by attempting them rather than by reading:

- **`library_ops.rs` ~38.** The four "component has empty name" checks can't fire on a library read from disk — the component name *is* the OLE storage name, so an empty one collides with the root and `save` refuses before a file exists. The adjacent `!is_finite()` checks can't fire either, since NaN saturates to 0.0 on write. There's a test pinning the first as impossible.
- **`server.rs` ~63.** ~34 inside its own test module, ~20 the async run loop (needs driving a real transport), ~9 a serialisation failure that cannot occur for a `Value`.
- **`maintenance.rs` ~10.** Need a `read_dir` failure, a path with no parent, or a non-UTF-8 filename.
- **`transport.rs` ~5** and **`tool_docs.rs` ~10** — the real stdin/stdout paths the test-sink bypasses, and assertion-failure branches.

That leaves ~390 genuinely reachable, of which ~115 would clear 99%. So the gate is attainable, but the remaining work is mechanical and scattered — the table-shaped wins are all spent. Happy to keep going; equally happy if you'd rather spend that effort on §A, or delete some of the unreachable defensive arms the way #383 did for the parsers.

## Checks

`cargo build`, `cargo test` (1,061 lib tests + every integration suite, 0 failures), `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings`. No production code touched. No CHANGELOG entry — nothing user-facing.


MatejGomboc (approved):
Full local verification: merged onto tonight's `main` (which now includes #386's byte-fidelity work your branch predates) — all 12 suites green, clippy `-D warnings` clean. The floor analysis is the most valuable part: ~125 confirmed-unreachable lines, established by *attempting* them rather than reading — that's now the documented reason the metric means what it means. The via serde-default coverage (zero thermal-relief gap = flood-connected, unsolderable board) is exactly the class of silent-wrong-value these tests exist for. Merge-ready; merge before #387.

PR #387, opened 2026-08-16
test: clear the 99% coverage gate

ande2407:
Closes the remaining gap to the 99% coverage gate (#302, TODO.md §B item 3).

**Builds on #385 — please merge that first.** This branch is cut from `test/close-the-coverage-gap`, so until #385 lands the diff here also shows its nine commits. Once #385 merges, this collapses to the two commits that are actually new: `001c83d` and `571739d`.

## Result

Measured with the CI command (`--ignore-filename-regex main.rs`):

| | line coverage | uncovered |
|---|---|---|
| #385 head | 98.94% | 439 |
| this PR | **99.01%** | **413** |

The gate allows 418 at the current 41,845 production lines, so this clears it by ~5 lines. Production code is unchanged — tests only, and no CHANGELOG entry for that reason.

## What is covered

- **Embedded-model paths end to end** — a `Models/Header` count disagreeing with the parsed index (the count is untrusted and must not drive the stream scan), an indexed model whose stream is absent, a payload that will not inflate, and the decompression-bomb cap driven through `decompress_capped`'s injectable limit rather than a 100 MB fixture.
- **Table-driven passes over the last two id tables** — every SchLib justification id, and every `PrimitiveKind` id/token pair, both including the unmapped-id fallback.
- **The SchLib Data-stream walk** through each framing fault it can meet: too short to hold a header, a zero-length end marker, a record claiming more bytes than remain, an unknown record type.
- **PcbLib FileHeader / Library/Data / Parameters** truncated at each stage of their framing, built with `cfb::create` so the damage is exact.
- **Sandbox rejection** on `render_footprint`, `render_symbol`, `extract_step_model` (library *and* output path) and `read_schlib` — none had one.
- **`include_geometry: false`** on `compare_components`, which had never been exercised: the flag defaults to `true`, so its false arm was dead in test. Plus the directional `only_in_a` arms, which a B-side-only fixture never reaches.
- `override_param`, `parse_key_order`, the base64 `None` arm, and the serde defaults no test had ever omitted.

## Two findings worth a look

**Fields inside `tracing::` macros only evaluate when a subscriber enables that level.** Calling the function was not enough — `parse_model_header_stream(&[])` left its guard line uncovered even under a direct test. Tests over damaged files now install a TRACE-level subscriber for their duration, which covers those lines and incidentally proves the diagnostics themselves format correctly.

**Three dead branches, all the guard-then-read shape #383 removed from the parsers.** Each was confirmed by writing a test that failed, not by reading:

- `query_update.rs` — both `update_*_component` reject an unknown name before the preview runs, so the `"component will be created"` arm cannot fire through the tool. Covered here by calling the `pub(crate)` preview helpers directly, which pins the contract but does not make the arm live.
- `pcblib/read_io.rs` — the enumerator only treats a storage as a component once a `Data` stream exists, so `read_footprint`'s missing-`Data` arm below it is unreachable.
- `schlib/read_io.rs` — the storage list is pre-filtered to names that have a `Data` stream, so the `open_stream` failure arm underneath cannot fire.

These look worth a follow-up PR in the same style as #383, together with the `parse_data_stream` dead arm already noted in #302. Left alone here to keep this change test-only.

## Gates

`cargo build`, `cargo test --all-features` (1102 lib tests + integration suites), `cargo fmt --all --check`, `cargo clippy --all-targets --all-features -- -D warnings` — all clean.


MatejGomboc (dismissed):
Verified the headline the hard way: merged this stack onto the **true** current `main` — which includes #386's byte-fidelity replay work, ~200 production lines your measurement predates — and re-ran the CI coverage command on the combined tree:

**99.00% — 417 uncovered of 41,845 production lines, gate allows 418.**

It survives the rebase by exactly one line. That's the honest number and it clears. Also: all 12 suites green and clippy clean on the combined tree, and I checked the "tests only" claim structurally — the three deletions are test-module import consolidations, and `decompress_capped`'s limit parameter already existed in production, so no seam was added.

The tracing-macro finding deserves calling out: fields inside `tracing::` macros only evaluating under an enabled subscriber is the kind of coverage phantom that would have wasted hours of someone's time later — now it's written down with the cure attached.

The three dead guard-then-read branches you confirmed by failing test: agreed on all three, and a follow-up in #383's style is welcome — `TODO.md` §B has your name next to that shape of work already.

With this merged, **the 99% coverage gate for v1.0.0 is met**. From your first question on #302 ("do you have a list of open items?") to clearing the release gate: #304–306, #379, #383, #385, #387. Genuinely outstanding work across the whole arc.

Merge order: #385 first, then this collapses to its two real commits. CI on this branch will run once that lands (or on maintainer approval, since it's a fork push).

PR #390, opened 2026-08-17
fix(mcp): replay all fidelity fields through the tool-layer JSON boundary

MatejGomboc:
Found during the pre-release review: the MCP tool layer silently dropped most of the byte-fidelity fields the readers emit, so an external MCP client doing read → write could not get the byte-identical output the library API guarantees. Regions already replayed theirs (`v7_layer`, `param_key_order`, `guid`) — which makes the rest an inconsistency inside one feature, not a design line.

## What was dropped

| Path | Field(s) | Symptom |
|------|----------|---------|
| `parse_pad` | `raw_tail` | pad extended tail rebuilt from template instead of replayed |
| `parse_via` | `raw_block`, `guid` | via block rebuilt; identity lost |
| `parse_text` | `raw_geometry` | allow-listed, then discarded |
| `parse_track` / `parse_fill` | `guid` | identity lost |
| `write_pcblib` | footprint `guid`, `primitive_order` | allow-listed, never read — kind-85 identity lost, primitives regrouped by kind |
| `write_schlib` | symbol `primitive_order` | same regrouping on the SchLib side |
| `read_pcblib` / `read_schlib` | footprint/symbol `guid` + `primitive_order` | never emitted at all (`get_component`, which serialises the whole struct, already emitted them — that is what the write allow-lists were extended for) |

## The fix

- Shared lenient `json_base64` helper (invalid base64 → absent + debug log, writer falls back to its template — the same Option-style leniency as the rest of the parser), also replacing the equivalent inline `image_data` decode.
- The five parser gaps wired through; both write tools now replay the footprint/symbol-level fields; both read tools emit them, mirroring the structs' serde shape (present only when carried).
- A malformed `primitive_order` is ignored, not fatal — it is advisory on the struct, and `write_sequence` is defensive against stale lists (pinned by test).

## Proof

`write_pcblib_replays_fidelity_fields_read_pcblib_emits` asserts **total JSON equality** across read → write → read with every fidelity field carried and a deliberately non-canonical primitive order (text, via, track, pad, fill). Because the equality includes the base64 `raw_tail`/`raw_block`/`raw_geometry` *re-read from the written file*, it proves the written binary blocks are byte-identical to the source's through the JSON boundary. The vacuity guard asserts the first read actually carries every replay field before comparing.

(That test also flushed out `write_pcblib`'s `.Designator` auto-injection — intentional and separately tested, so the fixture carries its own `.Designator` text as real footprints do. Whether that auto-add should become opt-in like the auto-3D-body after #182 is left as a separate decision.)

Gates: `cargo fmt` ✅, `clippy -D warnings` ✅, full suite ✅ (runs inside the instrumented coverage run), coverage on the production metric **99.01%** (416/42,074 uncovered — gate still met with the new lines).

### DIFF
--- a/src/altium/pcblib/mod.rs
+++ b/src/altium/pcblib/mod.rs
@@ -49,6 +49,7 @@ pub use primitives::{
     Arc, ComponentBody, DrillLayerPairType, EmbeddedModel, Fill, HoleShape, Layer,
     MaskExpansionMode, Model3D, Pad, PadShape, PadStackMode, PcbFlags, PowerPlaneConnectStyle,
     Region, RegionKind, StrokeFont, Text, TextJustification, TextKind, Track, Vertex, Via,
+    ViaStackMode,
 };
 
 use crate::altium::error::{AltiumError, AltiumResult};
--- a/src/mcp/tools/library_ops.rs
+++ b/src/mcp/tools/library_ops.rs
@@ -889,7 +889,7 @@ impl McpServer {
                             .collect()
                     };
 
-                    json!({
+                    let mut fp_json = json!({
                         "name": fp.name,
                         "description": fp.description,
                         "pads": pads,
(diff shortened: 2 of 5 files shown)


MatejGomboc:
Third commit completes the boundary sweep: **export_library** (both formats) and both **update_component** paths had the same drop class — export→import (the backup/restore loop) regrouped records and lost identities, and the get→update edit loop moved the designator back to its default position. Both now mirror the create path, pinned end-to-end against reopened files. Every remaining tool family mutates in-memory structs with no JSON boundary, so the sweep is complete. Coverage: **99.02%** (416/42,298).

PR #415, opened 2026-08-22
Potential fix for 1 code quality finding

MatejGomboc:
_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/embedded-society/altium-designer-mcp/security/quality/ai-findings)._

### DIFF
--- a/src/mcp/tools/maintenance.rs
+++ b/src/mcp/tools/maintenance.rs
@@ -721,6 +721,101 @@ impl McpServer {
             Err(e) => return ToolCallResult::error(e),
         };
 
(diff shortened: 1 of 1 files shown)


MatejGomboc:
Closing without merge. The finding is stale against current `main`: the `mechanical1…32` / `midlayer` / `internalplane` match this helper would replace does not exist (layer parsing goes through `Layer::parse`), so `parse_numbered_layer` would be dead code. The generated patch is also mangled by offset drift — it deletes `update_pad`'s save/result block and `call_update_primitive`'s signature while splicing a match arm into the middle of `update_pad` — and does not compile.

PR #416, opened 2026-08-22
Potential fixes for 2 code quality findings

MatejGomboc:
_This PR applies 2/2 suggestions from code quality [AI findings](https://github.com/embedded-society/altium-designer-mcp/security/quality/ai-findings)._


MatejGomboc:
Closing without merge. Both findings are non-applicable on current `main`: (1) `sanitise_file_name` already handles the all-underscores case, on the *trimmed* string — and its comment explains why the proposed `has_non_underscore` flag during the map pass is wrong (`"<>."` → `"__."` → trims to `"__"`, which the flag would wrongly accept); (2) `generate_unique_id` already uses `const ALPHABET` exactly as proposed. The generated patch is also mangled by offset drift (it drops `#[must_use]` and the function body, inserting statements inside a regex `get_or_init` closure) and does not compile.

PR #459, opened 2026-08-26
fix: build a library in memory and read and write the file once

MatejGomboc:
## Summary

Started as the follow-up to the flake @ande2407 hit on #458 (`perf_pcblib_save_100_footprints` at 2.0–2.8 s against a 2 s bound on a debug build). Redesigning that test as a **scaling** check exposed three real bugs behind the number — two of them inside a dependency.

**Bug #65 — saving was ~40× slower than it needed to be.** `save_atomic` handed the `cfb` writer a raw unbuffered `File`; a compound-file writer seeks and rewrites its FAT, directory and mini-stream tables on every stream it adds, and each was a disk round trip. The image is now built in memory and written to the temp file once, then renamed exactly as before. `restore_backup` shares the path.

**Bug #66 — opening had the mirror problem.** The readers seeked through an unbuffered `File` the same way. Both `open()`s now read the file into memory first.

**Bug #67 — the `cfb` crate itself is O(n²) in the number of entries**, on two counts: `MiniAllocator::seek_within_mini_sector` rebuilt the whole mini-stream FAT chain on every access to a small stream (the mechanism behind upstream issue mdsteele/rust-cfb#57), and every path lookup walked an unbalanced sibling tree (`insert_dir_entry` never rebalances — `// TODO: rebalance tree`). Fixed in **[mdsteele/rust-cfb#81](https://github.com/mdsteele/rust-cfb/pull/81)** (chains walked once and extended in place; a name index for lookups; **the on-disk tree untouched**, so bytes are identical) and pinned here through `[patch.crates-io]` until a release carries it.

Measured in a **release** build, 500 two-pad footprints:

| | before | after |
|---|---|---|
| `open()` | 135 ms (ratio 35× for 10× the footprints) | **20 ms** (8×) |
| `save()` | 1.75 s → 180 ms after #65 | **87 ms** (13×) |

The golden byte-identity suite (every fixture, every corpus twin) proves the written bytes are identical throughout.

**The perf tests now assert what they can prove.**
- **Scaling** — 500 footprints must cost ≤ 30× the 50-footprint time (linear ≈ 10, a quadratic ≈ 100); measured 8–15 in both build modes. Valid in any build — this is what caught #67.
- **Absolute bounds** — asserted only in an optimised build, printed otherwise; fastest-of-five timings.
- CI runs `cargo test --release --test perf_tests` after the existing release build.

## Verification

- fmt / clippy `-D warnings` clean; **1488 tests** green on the pinned crate
- Release perf: open 19 ms, save 87 ms; debug ratios 9 / 14; every bound met
- CHANGELOG `[Unreleased]` carries the lines

### DIFF
--- a/.github/workflows/ci_main.yml
+++ b/.github/workflows/ci_main.yml
@@ -171,6 +171,12 @@ jobs:
             - name: Run tests
               run: cargo test --locked --verbose
 
+            # The absolute timing bounds in tests/perf_tests.rs are asserted
+            # only in an optimised build (a debug build measures the build,
+            # not the code); the release artefacts above make this cheap.
+            - name: Run performance bounds (release)
+              run: cargo test --locked --release --test perf_tests -- --nocapture
+
             # Independent-reader check that generated files are Altium-readable
             # (issue #68). Install is best-effort; the harness skips cleanly if
             # pyaltiumlib is unavailable and only fails on a NEW regression.
--- a/.github/workflows/ci_pr.yml
+++ b/.github/workflows/ci_pr.yml
@@ -166,6 +166,12 @@ jobs:
             - name: Run tests
               run: cargo test --locked --verbose
 
+            # The absolute timing bounds in tests/perf_tests.rs are asserted
+            # only in an optimised build (a debug build measures the build,
+            # not the code); the release artefacts above make this cheap.
+            - name: Run performance bounds (release)
+              run: cargo test --locked --release --test perf_tests -- --nocapture
+
             # Independent-reader gate: confirm generated files are Altium-readable
             # (issue #68) by parsing them with pyaltiumlib. Dependencies are pinned
             # (requirements.txt) and isolated in a venv so the install is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   trip each time; the readers seeked through an unbuffered file the same way.
   A library is now built in memory and written once, and read into memory
   before it is parsed; the bytes are identical.
+- **Reading and writing a library now scale linearly with its size.** The
+  compound-file crate rebuilt the whole mini-stream chain on every access to
+  a small stream and walked an unbalanced directory tree on every path
+  lookup, so a library of `n` components cost a term in `n²`: 500 footprints
+  opened in 135 ms where 50 took 4 ms. Both are fixed in a patched `cfb`
+  the build pins until the fix is released upstream; 500 footprints now open
+  in 20 ms, and the bytes written are identical.
 - The performance tests assert what they can prove: that saving and opening
   scale linearly with library size (the accidental-quadratic guard, valid in
   any build), and absolute bounds only in an optimised build, which CI now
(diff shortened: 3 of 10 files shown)


ande2407:
Tested this on the machine whose flake prompted it (Windows 11, `--release`).

**The save path is fixed, and dramatically.** The whole `perf_tests` suite went
from ~29s to ~1.5s, and `pcblib_save_scales_linearly_with_footprint_count`
passes comfortably. The old `perf_pcblib_save_100_footprints` failed here on
every run at 2.0–2.8s against its 2s bound; that's gone.

**But the new open test fails consistently here — 8/8, not intermittently:**

```
20 footprints 3.42ms / 200 footprints 95.07ms (ratio 27.8)
20 footprints 3.48ms / 200 footprints 96.87ms (ratio 27.9)
20 footprints 3.67ms / 200 footprints 96.77ms (ratio 26.4)
20 footprints 3.44ms / 200 footprints 97.50ms (ratio 28.3)
20 footprints 3.67ms / 200 footprints 97.45ms (ratio 26.6)
20 footprints 3.54ms / 200 footprints 94.16ms (ratio 26.6)
20 footprints 3.48ms / 200 footprints 94.20ms (ratio 27.1)
20 footprints 3.48ms / 200 footprints 95.71ms (ratio 27.5)
```

Mean ratio ~27.3 against `MAX_SCALING_RATIO = 25.0`, spread 26.4–28.3. Since
`assert_scales` is fed `measure_min` over 5 runs, that tight a spread is not
noise — `PcbLib::open` really is mildly super-linear on this hardware.

Two things that argue the code is fine and the ceiling is just a shade tight:

- The absolute number is comfortable: 94–97ms for 200 footprints against the
  500ms `assert_bound`, which passes every time.
- Per the constant's own comment, "a quadratic would push the ratio towards
  100". 27 is nowhere near that — it reads as a real but modest super-linear
  term, not the regression the check exists to catch.

So this looks like a threshold call rather than a code problem. Happy either
way — raising the ceiling, or leaving it and treating my box as the outlier
that flushed it out. Shout if you'd like me to profile where the super-linear
term in `open` actually comes from; I have the harness set up.

MatejGomboc:
Thanks for running it on the box that started this — and your read was right: a tight 26–28× spread from min-of-five is a real super-linear term, not a threshold call. You were looking at the first head of the branch (20/200 footprints, ceiling 25); the commits since then found and fixed where that term lives.

It's inside the `cfb` crate, on two counts: `MiniAllocator::seek_within_mini_sector` rebuilds the whole mini-stream FAT chain on **every** access to a small stream (so with `n` component streams each read costs O(n)), and every path lookup walks a sibling tree that `insert_dir_entry` never rebalances (the `// TODO` there), so entries created in sorted order form a chain. Both are fixed in [mdsteele/rust-cfb#81](https://github.com/mdsteele/rust-cfb/pull/81) — chains walked once and extended in place, plus a name index for lookups, with the on-disk tree untouched so our bytes stay identical — and the build pins that commit through `[patch.crates-io]` until a release carries it.

On the current head (`baad020`) the test compares 50 vs 500 footprints with a ceiling of 30, and measures here at **open 8.0×, save 12.6×** in release (9 / 14 in debug); 500 footprints open in ~20 ms where the old crate took 135 ms. If you can re-run `cargo test --release --test perf_tests -- --nocapture` at that head I'd value the numbers from your hardware — it's the machine that made the check bite in the first place.

No profiling needed now, but thank you for the offer; the harness you have set up may be worth pointing at `SchLib` next, which goes through the same crate paths.

PR #460, opened 2026-08-28
fix: refuse a description Altium will not import

ande2407:
## The bug

Altium accepts a component description of any length through the file format, then **fails to import the whole library** if one exceeds 256 characters — naming neither the library nor the offending component. So a library written by this server could be silently unusable, and the only symptom was a failed import with nothing to go on.

Found in the field: a generated 62-part library had 12 footprint descriptions between 257 and 326 characters. `write_pcblib` reported `status: success` with `validation: {"error_count": 0, "warning_count": 0}`, and the library would not import.

## The fix

**Refuse it at the API boundary.** `write_pcblib` and `write_schlib` now reject an over-length `description` before anything is written:

```
Footprint 'MHS-JA0' has a 307-character description, but Altium will not import
a library whose description exceeds 256 characters. Shorten it by 51 characters.
```

The error names the component and the overshoot, so it is fixable in one edit. Covers all three description fields that reach a component record: footprint (`write_pcblib`), symbol (`write_schlib`), and the footprint link inside a symbol.

This is a hard error rather than a validation warning on purpose. Post-write validation runs *after* the file exists, so a warning there would still leave an unopenable library on disk.

**Also caught by the validators.** `validate_library` and the post-write validation now report the same finding as `severity: "error"`. That covers libraries this server did not author — an older build, or a description carried in verbatim by `copy_component` / `copy_component_cross_library` / `import_library`, which do not go through the parse path.

## Notes on the limit

- Counted in **characters**, matching how Altium states it. For the ASCII text these fields almost always hold, that is also the byte count. If the real constraint turns out to be bytes, `check_description_len` is the one place to change.
- Pin and parameter descriptions are **not** covered. They are separate fields and I have no evidence of a 256-character limit on them; adding a guess seemed worse than leaving them.

## Tests

Five new tests in `read_write::tests::failure_paths`:

- a description of exactly 256 characters is accepted and round-trips
- 257 characters is refused, the error names the footprint, states the limit and the overshoot, **and no file is created**
- an over-length symbol description is refused
- an over-length footprint-link description is refused
- `validate_library` reports an over-length description as an `error`, naming the component — built through the library API to sidestep the new parse-time refusal

`cargo test --lib`: 1239 passed, 0 failed. `cargo clippy --all-targets`: clean.

## Docs

- `CHANGELOG.md` under Unreleased/Fixed.
- `docs/AGENT_GUIDE.md` gains a short "Description length" section. That file ships in the `initialize` response, so binary-only users get the rule without reading the source.
- The limit is stated on the `description` fields in the tool schemas, so it appears in `tools/list` and in the generated `docs/TOOLS.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A description longer than Altium will import is now refused instead of
+  written.** Altium silently accepts a component description of any length
+  through the file format but then fails to import the whole library if one
+  exceeds 256 characters, naming neither the library nor the component. So
+  `write_pcblib` and `write_schlib` now reject an over-length `description` at
+  the API boundary, before anything is written, and say which component it is
+  and by how many characters to shorten it. This applies to footprint, symbol
+  and footprint-link descriptions. `validate_library` reports the same thing as
+  an `error` on libraries this server did not author — an older build, or a
+  description copied in from elsewhere — and so does the post-write validation
+  that `copy_component` and `import_library` pass through.
+
 - **A solid symbol body no longer paints over the pin names inside it.**
   `write_schlib` recorded the order it happened to parse its input in — pins
   before rectangles — and replayed that as if it were an authoring order, so a
--- a/docs/AGENT_GUIDE.md
+++ b/docs/AGENT_GUIDE.md
@@ -52,6 +52,15 @@ behind their pin names. Keep bodies solid — a transparent body is not the fix
 for hidden pin names. Supply `primitive_order` only to reproduce an order you
 read off an existing file; it overrides the default.
 
+## Description length
+
+Altium will not import a library if any component description exceeds **256
+characters** — and it reports neither which library nor which component is at
+fault. `write_pcblib` and `write_schlib` therefore refuse an over-length
+`description` before writing anything, naming the component and the overshoot.
+Keep footprint, symbol and footprint-link descriptions inside the limit; put
+the long-form reasoning in your own notes, not the description field.
+
 ## Filesystem sandbox
 
 Only paths inside the configured `allowed_paths` (see `config.json`) can be read
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -923,7 +923,7 @@ Manage footprint links in Altium SchLib symbols. Supports listing, adding, and r
 | Name | Type | Required | Description |
 | --- | --- | --- | --- |
 | `component_name` | string | yes | Name of the symbol to manage footprints for |
-| `description` | string | no | Footprint description (optional for add) |
+| `description` | string | no | Footprint description (optional for add). Max 256 characters — Altium will not import a library with a longer description, so an over-length value is refused. |
 | `filepath` | string | yes | Path to the Altium SchLib file |
 | `footprint_name` | string | no | Footprint name (required for add, remove) |
 | `library_path` | string | no | Optional (add): absolute path to the .PcbLib containing the footprint, written as ModelDatafile0 so Altium can resolve and preview the model. Omit to link by name only (requires the library to be installed/in the project, else 'footprint not found'). |
--- a/src/mcp/tool_definitions.rs
+++ b/src/mcp/tool_definitions.rs
@@ -231,7 +231,7 @@ impl McpServer {
                                     },
                                     "description": {
                                         "type": "string",
-                                        "description": "Footprint description"
+                                        "description": "Footprint description. Max 256 characters — Altium will not import a library with a longer description, so an over-length value is refused."
                                     },
                                     "pads": {
                                         "type": "array",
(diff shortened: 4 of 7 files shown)


MatejGomboc (approved):

MatejGomboc:
Reviewed on a local checkout of d7b0176 (based on current main): fmt, clippy `-D warnings`, the full suite (**1494 tests**) and both corpus sweeps are green — no real library has a description anywhere near the limit (the longest across ~300 Altium-authored components is 91 characters), so replay is unaffected. The code is tidy: one constant, one check, all four validators sharing one message, and the parse-boundary placement means `update_component` gets it for free. Thank you for catching a failure that gives the user nothing to go on.

Before merging I wanted to know where the limit actually lives, so I measured it against Altium itself. Five PcbLibs written through the library API with descriptions of **200, 255, 256, 257 and 300** characters, opened in a headless AD24 by `scripts/Verify-Libraries.ps1` (which I've extended to report description lengths as Altium reads them):

```
desc_200.PcbLib  opened  components=[DESC_200]  description_lengths=[200]
desc_255.PcbLib  opened  components=[DESC_255]  description_lengths=[255]
desc_256.PcbLib  opened  components=[DESC_256]  description_lengths=[256]
desc_257.PcbLib  opened  components=[DESC_257]  description_lengths=[257]
desc_300.PcbLib  opened  components=[DESC_300]  description_lengths=[300]
```

Altium Designer opens every one and reads the full description back, 300 characters included. So the file format has no limit, and neither does Altium's PcbLib reader; the rejection you saw belongs to the **workspace / library-import path** (the docs cap *Designator* and *Comment* at 255 and say nothing about Description). For a file-based user — which is most of ours, and the workflow the README describes — a 300-character description works.

That argues for one change: keep every check, but make the finding a **validation `warning`** (in `validate_library` and post-write validation, naming the component and the overshoot, and saying *which* Altium path refuses it) rather than refusing the write. Refusing would make the server stricter than Altium is for the common case, while the warning gives the workspace user exactly the clue the importer withholds — and post-write validation already runs on the copy/import paths, so the three write paths that skip the parse-time check today (`manage_schlib_footprints`, the `description` argument of `copy_component` / `copy_component_cross_library`, and `import_library`) are covered too. The schema notes and the AGENT_GUIDE section stay, reworded to name the import path.

Two questions where you have evidence I don't:
1. Was the failing import the Altium 365 workspace importer, or something else (an IntLib compile, a database library)? If an IntLib compile also fails, the hard refusal has a much stronger case, since that path is common.
2. Did a description of exactly 256 characters import, or is 256 the first length you *didn't* see fail? Altium's documented caps are 255, so the boundary may sit one lower.

Happy to make the warning change on your branch if you'd rather not — say the word.

ande2407:
The error was in the library importer, specifically complaining about the
footprint description being longer than 256 characters. I can snag a
screenshot when I'm back on that computer. It gave that error, and denied
the import to Altium 365. We can go either way, but it should get at least
a warning before you get to the step of not being able to be imported into
the system.

On Saturday, Aug 29, 2026, 6:53 AM, wrote:

MatejGomboc left a comment (embedded-society/altium-designer-mcp#460)
Reviewed on a local checkout of d7b0176 (based on current main): fmt,
clippy -D warnings, the full suite (1494 tests) and both corpus sweeps are
green — no real library has a description anywhere near the limit (the
longest across ~300 Altium-authored components is 91 characters), so replay
is unaffected. The code is tidy: one constant, one check, all four
validators sharing one message, and the parse-boundary placement means
update_component gets it for free. Thank you for catching a failure that
gives the user nothing to go on.
Before merging I wanted to know where the limit actually lives, so I
measured it against Altium itself. Five PcbLibs written through the library
API with descriptions of 200, 255, 256, 257 and 300 characters, opened in a
headless AD24 by scripts/Verify-Libraries.ps1 (which I've extended to
report description lengths as Altium reads them):
desc_200.PcbLib opened components=[DESC_200] description_lengths=[200]
desc_255.PcbLib opened components=[DESC_255] description_lengths=[255]
desc_256.PcbLib opened components=[DESC_256] description_lengths=[256]
desc_257.PcbLib opened components=[DESC_257] description_lengths=[257]
desc_300.PcbLib opened components=[DESC_300] description_lengths=[300]

Altium Designer opens every one and reads the full description back, 300
characters included. So the file format has no limit, and neither does
Altium's PcbLib reader; the rejection you saw belongs to the workspace /
library-import path (the docs cap Designator and Comment at 255 and say
nothing about Description). For a file-based user — which is most of ours,
and the workflow the README describes — a 300-character description works.
That argues for one change: keep every check, but make the finding a
validation warning (in validate_library and post-write validation, naming
the component and the overshoot, and saying which Altium path refuses it)
rather than refusing the write. Refusing would make the server stricter
than Altium is for the common case, while the warning gives the workspace
user exactly the clue the importer withholds — and post-write validation
already runs on the copy/import paths, so the three write paths that skip
the parse-time check today (manage_schlib_footprints, the description
argument of copy_component / copy_component_cross_library, and
import_library) are covered too. The schema notes and the AGENT_GUIDE
section stay, reworded to name the import path.
Two questions where you have evidence I don't:

Was the failing import the Altium 365 workspace importer, or something else
(an IntLib compile, a database library)? If an IntLib compile also fails,
the hard refusal has a much stronger case, since that path is common.
Did a description of exactly 256 characters import, or is 256 the first
length you didn't see fail? Altium's documented caps are 255, so the
boundary may sit one lower.

Happy to make the warning change on your branch if you'd rather not — say
the word.
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull
requests on the go with GitHub Mobile for iOS and Android. Download it
today!

You are receiving this because you authored the thread.Message ID:
***@***.***>

MatejGomboc:
Thanks — that settles it. The Altium 365 library importer is the one path that refuses, its own message says 256, and Altium Designer itself opens and reads longer descriptions fine, so the right shape is exactly what you said: a warning the user sees *before* the import step, not a refusal at write time. Follow-up PR coming: the checks and the 256 constant stay, the write-time refusal becomes a post-write validation warning (so it also reaches the copy/import/link paths), and every message names the Altium 365 importer as the reason. A screenshot would still be welcome on that thread for the record, but nothing hinges on it.

PR #497, opened 2026-09-03
feat: a Dockerfile that builds the same --locked release into a minimal image

MatejGomboc:
Adds a container route (prompted by the awesome-mcp-servers bot, which now requires a Glama listing with a working Dockerfile before merging #13495 — but useful on its own for a Linux box, a NAS or a CI job that generates libraries into a mounted folder).

- **`Dockerfile`**: `rust:1.95.0-slim-bookworm` builder running the same `cargo build --locked --release` as the published binaries → `debian:bookworm-slim` runtime, unprivileged `mcp` user, `VOLUME /libraries`, `ENTRYPOINT altium-designer-mcp`, `CMD --allow /libraries` (the mount is the whole allow-list). Same Debian release in both stages so glibc matches.
- **`.dockerignore`**: allow-list of what the build needs — `src/`, `Cargo.*`, `LICENCE`, and `docs/AGENT_GUIDE.md` (it's `include_str!`-ed into the server); `rust-toolchain.toml` deliberately excluded so rustup doesn't install rustfmt/clippy inside the build.
- **`check-toolchain-pin.sh`** now also holds the Dockerfile's `FROM rust:X.Y.Z` to `rust-toolchain.toml` — bumping Rust means bumping all three.
- **Dependabot**: `docker` ecosystem in the weekly batch for the base images.
- **README** § Installation: the container paragraph; **CHANGELOG** Added.

Verified locally: image builds (124 MB, ~3 min), `--version` → `altium-designer-mcp 1.0.1`, MCP `initialize` OK, `tools/list` → 34 tools with titles, runs as `uid=10001(mcp)`.

After merge, the remaining awesome-list steps are: claim the Glama listing and add the Dockerfile there (owner UI step), then add Glama's score badge to the PR entry.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+# Allow-list: the image build needs only the crate sources. Everything else —
+# target/, node_modules/, the golden fixtures, the Altium scripts, .git — would
+# bloat the build context without changing the binary.
+*
+!Cargo.toml
+!Cargo.lock
+!src/
+!docs/AGENT_GUIDE.md
+!LICENCE
+
+# rust-toolchain.toml is deliberately NOT sent: the builder image already
+# carries the pinned compiler, and the file would make rustup try to install
+# its listed components (rustfmt, clippy) at build time for no benefit.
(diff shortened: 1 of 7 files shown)


MatejGomboc:
Added `glama.json` (`maintainers: ["MatejGomboc"]`) — Glama's own methodology: the repo carries the Dockerfile it builds and a `glama.json` naming who may claim the listing; it re-indexes within minutes of a push.

